PR #322, opened 2018-03-02
Add a Dockerfile which contains a usable dev and build environment

theatrus:
The resultant image can be used to iterate on the code base, especiall
on the front end code. Installs required NPM and Node versions, as
well as installing Go 1.10.

The Dockerfile is not used for Travis CI yet, as it requires a Ruby
install for the actual packaging work. I didn't want to lump over the
entire CI and test process in one shot.

The container doesn't currently "run" the service, though this could
be done. Its expected to be run in a shell environment at this stage.

Signed-off-by: Yann Ramin <atrus@stackworks.net>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+LABEL maintainer="code@reef-pi.com"
+
+# Use debian as it matches a Raspbian environment
+FROM debian:stretch
+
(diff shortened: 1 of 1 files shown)


ranjib:
🤗

Issue #468, opened 2018-07-25
Allow sending test email as part of telemetry alert configuration


ranjib:
Users find it hard to test if they have setup email alerting capability correctly or not. Having a "send test alert" button will address this

PR #487, opened 2018-08-11
(feat) test email

ranjib:


ranjib:
https://github.com/reef-pi/reef-pi/issues/468

Issue #464, opened 2018-07-24
Change all mention of equipments in codebase


ranjib:
equipments is not a correct English word and have been a constant source of annoyance in different threads. since the same word is used by 1.0 API (/api/equipments), lets use 2.0 as an opportunity to get rid of all usage of equipments in reef-pi in lieu of equipment.

PR #492, opened 2018-08-14
(chore)nuke all reference of equipments

ranjib:


ranjib:
https://github.com/reef-pi/reef-pi/issues/464

PR #512, opened 2018-09-01
Cleaning Build environment for UI

zekth:
As we discussed i've done some modifications and refactoring in the UI. Regarding the build process i also i've modified the api.go to aim the 'ui' folder which is the build folder of webpack to correctly serve the rendered files. Added to this i've also modified makefile for refactoring purpose and also fit the new ui folder.

Also i've added support of sass, so all the specific styling has to be registered in the Style.scss via import and not in the react app direclty. By this way we ensure compiling all the css of the app in one file.

Added also editorconfig to fit the IDE configuration between developers.

With all these modifications you'll have to do a `npm install` to resolve the new packages.


jsx/main_panel.jsx
```diff
@@ -15,7 +15,7 @@ import {fetchUIData} from 'redux/actions/ui'
 import {fetchInfo} from 'redux/actions/info'
 import {connect} from 'react-redux'
 import Summary from 'summary'
-import '../assets/reef_pi.css'
+// import '../assets/reef_pi.css'
```
ranjib: delete this line

package.json
```diff
@@ -30,31 +47,22 @@
...
+    "webpack": "^4.10.2",
     "webpack-cli": "^2.1.4",
+    "webpack-dev-server": "^3.1.4",
     "webpack-livereload-plugin": "^2.1.1",
-    "webpack": "^4.10.2",
-    "webpack-dev-server": "^3.1.4"
+    "yup": "^0.26.2"
   },
   "scripts": {
+    "build": "./node_modules/.bin/webpack -d",
```
ranjib: do you need -d  flag?

zekth: It's a debug flag. We don't "need" it but for devenv it's easier for debugging. When we build for release we don't use this flag

jsx/main_panel.jsx
zekth: Seems legit

Issue #503, opened 2018-08-22
Invalid Login Feedback


Ranthalion:
The system should provide feedback upon invalid receiving invalid credentials.  The system currently proceeds as though credentials were valid, but shows confusing errors and blank pages.

ranjib:
Agree, the UI side implementation of auth is in general bad, it uses localstore and I think its not as par best practices.

zekth:
I'll improve this part.

PR #531, opened 2018-09-08
Authentication based on cookies

zekth:
Added session authentication and UI handling. This will fix:
https://github.com/reef-pi/reef-pi/issues/503

ATM it's based on /api/errors to check if authenticated or not. Maybe could add a path /api/me ?


controller/reef_pi.go
```diff
@@ -158,3 +163,29 @@ func (r *ReefPi) Controller() types.Controller {
...
+// type Auth struct {
+// 	user string
+// 	pass string
+// }
+
+// func (r *ReefPi) check(user, pass string) bool {
+// 	return (a.user == user) && (a.pass == pass)
+// }
+
+func (r *ReefPi) BasicAuth(fn http.HandlerFunc) http.HandlerFunc {
```
ranjib: delete ^^

jsx/app.jsx
```diff
@@ -20,5 +32,17 @@ export default class App extends React.Component {
         </div>
       </div>
     )
+  }else{
+    return (<SignIn />)
+  }
+}
```
ranjib: instead, 
```js
    if(!this.state.logged){
       return (<SignIn />)
    }
    return (
      <div className='container'>
        <div id='reef-pi-alert' />
         ....
         </div>
      </div>
    )
  }
```

PR #536, opened 2018-09-11
(chore)fix lint violations

ranjib:


ranjib:
Incorporated in https://github.com/reef-pi/reef-pi/pull/537

PR #540, opened 2018-09-12
UI improvement

zekth:
Improved UI + Fix SignOut method


Ranthalion:
:+1:

assets/sass/style.scss
```diff
@@ -7,6 +7,18 @@
...
+.pointer {
+  cursor: pointer;
+}
+
+.bottom-bar {
+  border-top: solid 1px #CCC;
+}
+
+.custom-select {
+  background: #FFF;
```
Ranthalion: I found this difference between the bootstrap site styline and reef-pi styling for .custom-select. 
-webkit-appearance: none;

I don't think we should need to change with background: #FFF; I would prefer to find root cause of losing -webkit-appearance:none

Ranthalion: Based on https://getbootstrap.com/docs/4.1/getting-started/webpack/ I think it needs postcss-loader

ranjib: @Ranthalion merging this for the time being, if you find this is problematic or annoying, happy to revisit this

zekth: I'll add post css loader so.

Issue #451, opened 2018-07-14
Adafruit.io AIO Key Not Saving


spamr:
Upon entering my Adafruit.io username I receive the following message instantly "Please set a valid adafruit.io key" and after hitting "update" the key does not save. 

When reloading the page after saving the AIO key the textbox is null.

running 2.0.0-beta-6

ranjib:
this is fixed in master, will be part of next release. Couple of notes:
- reef-pi UI will never show existing/set password & token. hence the adafruit.io token field is empty , thats not a bug. It does not mean reef-pi is not save the new token when user type it in the field and hit  save button. We'll add information text message in the UI to make this clear
- I have ensured 2.0 sensors (ph, temperature and ato) along with their usage (heater/cooler) are all integrated with adafruit.io based metric emission

Let us know if you have any concern,

PR #544, opened 2018-09-14
(bug)integrate adafruit.io with ph subsystem

ranjib:


ranjib:
https://github.com/reef-pi/reef-pi/issues/451

PR #550, opened 2018-09-15
Win support

### LINKED ISSUE #543: Support conditional compilation for windows
Add support for development on windows based machines using conditional compilation.

Ranthalion:
Update .gitignore to include .lock
Add cross-env to support cross platform env vars
Update standard config and fix lint
Add wmi dependency to Makefile
Add windows specific commands to Makefile
Refactor to abstract OS specific functionality.
fixes #543


Ranthalion:
go build seems to be ignoring the conditional build comment in daemon_windows.go...  I don't understand the problem, so this will probably be on hold for a few days.

commands/daemon_windows.go:11:19: daemonize redeclared in this block
	previous declaration at commands/daemon_notwin.go:14:19

Ranthalion:
Identified root cause of build error in makefile.  Corrected the issue and this PR is ready for review and can be merged.

Makefile
```diff
@@ -57,6 +61,23 @@ go-get:
...
+	go get -u github.com/gorilla/sessions
+	go get -u gopkg.in/robfig/cron.v2
+	go get -u github.com/dustin/go-humanize
+	go get -u github.com/reef-pi/rpi/pwm
+	go get -u github.com/reef-pi/drivers
+	go get -u github.com/reef-pi/adafruitio
+	go get -u github.com/nfnt/resize
+	go get -u github.com/coreos/bbolt
+	go get -u github.com/StackExchange/wmi
+
```
ranjib: @Ranthalion I think we can keep the common bits in a common make target, and have the windows task depend on that, and explicitly calls out only go get stackexchange/wmi , that will remove some duplication. What do you think?

commands/daemon_notwin.go
```diff
@@ -1,39 +1,42 @@
-package main
```
ranjib: lets call it daemon_linux.go instead

ranjib:
Looks good to me, just have couple of minor nits,

commands/daemon_notwin.go
Ranthalion: Agreed on both parts. I'll make revisions today.

Ranthalion:
consolidated makefile.  daemon_notwin.go was renamed to daemon.go.  daemon_linux.go caused it to be skipped on MacOS.

PR #558, opened 2018-09-22
Update connectors to use id as the key instead of index. Fixes #555

### LINKED ISSUE #555: Outlet deletion gives inconsistent list
Steps to reproduce:
- Create 8 outlets.
- Delete an outlet from top of the list
- Notice the outlet remains on the list, and outlet below is shown as deleted.
- Refresh shows correct outlet list. This is UI only bug, i.e. in the backend, the proper outlet is deleted
- This does not happen if outlets from the bottom of the list are deleted

Ranthalion:


ranjib:
🍠

PR #559, opened 2018-09-22
Unique indexes

zekth:
Fixing the errors with unique index on react rendering


Ranthalion:
Consider using fragments instead of arrays.

Ranthalion:
:+1:

PR #567, opened 2018-09-24
Unit tests

zekth:
Improving unit tests and Coverage


jsx/utils/confirm.js
```diff
@@ -3,14 +3,14 @@ import React from 'react'
...
-export function confirm (message, options) {
+export function confirm (message, options = {}) {
   var cleanup, component, props, wrapper
-  if (options == null) {
-    options = {}
-  }
-  props = $.extend({
-    message: message
-  }, options)
+  props = $.extend(
```
ranjib: is this doable without jquery?

PR #574, opened 2018-09-30
Ui timer

Ranthalion:
- Timer UI improved to use responsive design and include UI validations
- Refactor to move shared ui components into shared folder
- Fix issue with creating new ATO not defaulting status
- Fix issue with creating new Temperature not defaulting values correctly
- Add CollapsibleList and Collapsible UI components.  Used in Timer, pending refactor to other tabs

Desktop
image
 
Mobile
image


jsx/ato/edit_ato.jsx
```diff
@@ -4,7 +4,7 @@ import { ErrorFor, ShowError } from '../utils/validation_helper'
 import { showAlert, clearAlert } from 'utils/alert'
 import classNames from 'classnames'
 import { Field } from 'formik'
-import BooleanSelect from '../temperature/boolean_select'
+import BooleanSelect from '../ui_components/boolean_select'
```
zekth: So migration of common components will be in /ui_components? sounds nice
May we add all common buttons in it? like add/delete/edit ?

jsx/ato/ato_form.jsx
```diff
@@ -4,17 +4,26 @@ import { withFormik } from 'formik'
...
-  }),
+  mapPropsToValues: props => {
+    let data = props.data
+    if (data === undefined) {
+      data = {
+        enable: true,
+        notify: {}
+      }
+    }
+    return {
```
zekth: Instead of doing this kind of checking would it be better to use a Class and call a Sanitize method in it? would be more readable in my opinion.

jsx/ato/edit_ato.jsx
Ranthalion: Delete and Edit were integrated into the collapsible list, which probably should be renamed.  I think it makes sense to consider making components from other common input patterns.  I also think the duration inputs are a candidate to be turned into a component and reused.

zekth: I think an inventory of those components and refactor would really help us for further developments. Nice initiativ!

jsx/ato/ato_form.jsx
Ranthalion: This is specifically to handle a null input object.  I would prefer to use safe navigation operator instead of this, but it is not available in the version of babel we are using.  After 2.0 release, I would like to update babel to get access to the newer features.  Then I can replace this with something more like 
```
notify: props?.data?.notify?.enable ?? false,
```

zekth: This is really weird for me because we are the master of the Data. So in this case we admit the data may be corrupted in a sense. After 2.0 it could be nice to check if we want to use strong types like we can in TypeScript for example

PR #583, opened 2018-10-03
Log center

zekth:
Refactoring notifications and added Log Center.

Also fixed windows env tweeks + webpack optimization


ranjib:
@zekth i could not spot any reason for modeling this as capabilities (i.e. we dont have any backend API for this feature), if its all in the UI, we can skip modeling this as capability and make it availables to all users (i.e. they dont have to opt in for this)

zekth:
> @zekth i could not spot any reason for modeling this as capabilities (i.e. we dont have any backend API for this feature), if its all in the UI, we can skip modeling this as capability and make it availables to all users (i.e. they dont have to opt in for this)

Basically it was to fit the MainPanel tab behavior to appear in the menu.

ranjib:
We can skip the backend parts (go) and just add this tab explicitly in the ui I think, similar to the configure tab

zekth:
> We can skip the backend parts (go) and just add this tab explicitly in the ui I think, similar to the configure tab

OK for me. Anything else?

ranjib:
reset the go files, we dont need any changes on the backend for this PR, after that this is good to merge

PR #584, opened 2018-10-04
Improve Connectors

zekth:
Improve Connectors panel,
deleting all calls to jQuery

Improving unit tests. Still needs some validations and Ui refactor

### DIFF
--- a/jsx/connectors/jack.jsx
+++ b/jsx/connectors/jack.jsx
@@ -77,7 +77,7 @@ export default class Jack extends React.Component {
               type='text'
               id={'jack-' + this.props.jack_id + '-name'}
               onChange={this.handleNameChange}
-              className='form-control'
+              className='jack-name form-control'
               value={this.state.name}
             />
           </div>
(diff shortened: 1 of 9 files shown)


jsx/connectors/jack.jsx
```diff
@@ -151,6 +154,6 @@ Jack.propTypes = {
   name: PropTypes.string.isRequired,
   pins: PropTypes.array.isRequired,
   jack_id: PropTypes.string.isRequired,
-  remove: PropTypes.func.isRequired,
+  remove: PropTypes.func,
```
Ranthalion: I'd rather not remove app code to facilitate tests.  Try wrapping jest.fn() with a function for test coverage. 
See https://github.com/facebook/jest/issues/6329

Ranthalion:
:shipit:

jsx/connectors/jack.jsx
zekth: was a forget. Fixed the problem and found whyt this was not working.

Issue #590, opened 2018-10-08
Migrating deprecated ShowAlert


zekth:
In https://github.com/reef-pi/reef-pi/commit/0eb5557021dad432021462cf014c7bfbd853e6e6 the ShowAlert was deprecated. A fallback has been made but it shows console.warn. It would be great to clean those deprectated methods before v2 release.

ranjib:
+1

Ranthalion:
There are some cases where the system should show an alert to the user, but not log it in the notification center.  A specific use case is when the user attempts to submit a form with validation failures.  An alert is used to notify the user that the data was not submit.  In this case, it is not a system error or an event that needs to be tracked; it's simply a communication mechanism.

utils/alert could be refactored to provide convenience methods for communicating with the user and wrapping the dispatch to log.  I'm not really a fan of boolean flags in method parameters to control functionality, but this could be the start of a proposal.
```
utils/alert {
  showInfo(message: string, addToLog: bool)
  showWarning(message: string, addToLog: bool)
  showError(message: string, addToLog: bool)
}
```

zekth:
what about:
```
utils/alert {
  showInfo(message: string) // no ui Log
  showWarning(message: string) // no ui Log
  showError(message: string) // no ui Log
  logInfo(message: string) // Alert and Log
  logWarning(message: string) // Alert and Log
  logError(message: string) // Alert and Log
}
```

PR #591, opened 2018-10-08
Form Validation and responsiveness for Doser

### LINKED ISSUE #579: Update Doser UI
Add UI validation to Doser tab and update to responsive design.

Ranthalion:
- Add validation and responsiveness to Doser UI
- Update Confirm's cancel button style
- Move doser calibration to modal
- Tweak Collapsible to stay expanded during edit mode
- Updated Doser API functionality to maintain cron schedule during Create and Update.
Fixes #579 

Preview available at http://reef-pi.herokuapp.com

image
image
image


jsx/doser/calibrate.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import * as Yup from 'yup'
+import { ErrorFor, ShowError } from '../utils/validation_helper'
+import classNames from 'classnames'
+import { withFormik, Field } from 'formik'
+
+export const Calibrate = ({
```
zekth: Why not using the class pattern for component like everywhere in the UI? Makes it more consistent.

jsx/doser/edit_doser.jsx
```diff
@@ -0,0 +1,220 @@
...
+  readOnly
+}) => {
+  const handleSubmit = event => {
+    event.preventDefault()
+    clearAlert()
+    if (dirty === false || isValid === true) {
+      submitForm()
+    } else {
+      submitForm() // Calling submit form in order to show validation errors
+      showAlert(
```
zekth: Use the dispatch nofication because this call is deprecated

ranjib: Do you mean addLog ? The main panel has notification component already

jsx/doser/calibrate.jsx
ranjib: It’s ok to have mixed style if we don’t have any state based logic, the react style guide recommends this as well. I remember asking the same to @Ranthalion

Ranthalion: In react, it is preferred to use stateless function if the component does not need to track state.  Some linters even include this as a rule, but the one we are using doesn't seem to include it.

jsx/doser/edit_doser.jsx
Ranthalion: The dispatch call is pretty long.  What about keeping showAlert as a convenience method instead?  This also removes the need to rewrite a bunch of the other code that is already using showAlert.

Ranthalion: This also places an entry in the notification log.  In this case, the alert is specifically to communicate with the user; it should not be persisted in the log.  It is not a system error, it is a validation failure.
Maybe the right answer is to refactor utils/alert to provide functionality that makes it easy to show info/warning/error and log to notification as appropriate.  I'll think on this and add comments to the issue posted earlier.

zekth: https://github.com/reef-pi/reef-pi/blob/master/jsx/utils/ajax.js#L27 like here but there is a method to log only and no display. May i add an entry to show notification and don't log it?

jsx/doser/calibrate.jsx
zekth: So do you suggest me to migrate in this pattern for Notification Alerts?

Ranthalion: I don't think it's a big deal.  I quickly looked for any benchmarks that show performance differences, but didn't really find anything compelling.  
Functional components vs Stateful components was a topic when I went to a Reactjs meetup.

jsx/doser/edit_doser.jsx
Ranthalion: See Issue #590 for new comments and additional conversation.

PR #592, opened 2018-10-09
Refactor Log / Alert behavior

zekth:
Refactor of the Log / alert components to have a better handling of all the use cases.
This give use the flexibility to show notifications to user without logging it and so on. Like @Ranthalion mentionned in https://github.com/reef-pi/reef-pi/issues/590


Ranthalion:
Looks good.
:shipit:

PR #600, opened 2018-10-14
Form Validation and Responsiveness for Macro

Ranthalion:
Preview available at http://reef-pi.herokuapp.com

- Add validation and responsiveness to Macro UI
- Add responsiveness to alerts
- Add url parameter to smoke test
Fixes #580 

image

image

PR #601, opened 2018-10-14
Revert "Form Validation and Responsiveness for Macro"

ranjib:
Reverts reef-pi/reef-pi#600


ranjib:
@zekth @Ranthalion we wont need this right? we are good with the current implementation.

PR #624, opened 2018-10-27
(chore)Ignore api error when issuing reload

ranjib:


ranjib:
@zekth review please

ranjib:
its instant (less than a second), so I expect this will not be noticeable. If reef-pi has failed then the connection error will prop up with troubleshooting details

PR #630, opened 2018-11-03
(bug)fix diurnal config behavior when time range crosses midnight

ranjib:


Ranthalion:
👍 
Had to debug to understand what this was fixing.  Nice.

ranjib:
I think theres still a bug :-( . I am adding more asserts.

PR #633, opened 2018-11-04
Fix component-selector to appear over footer

### LINKED ISSUE #632: Unable to configure bottom dashboard entry
From Erica-Renee
"When in Dashboard configuration. the Browser (using chrome ) Will not let you see the bottom of the Dialog box to select dashboard item Just not sure anyone has seen this.. Its only a issue adding temp to bottom of dashboard"

Ranthalion:
Change z-index for div.drop-down menu.  This will set all button based dropdowns to appear above fixed footer.
Fixes #632 
image

### DIFF
--- a/assets/sass/style.scss
+++ b/assets/sass/style.scss
@@ -39,6 +39,10 @@ html {
   margin-bottom: 10px;
 }
 
+div.dropdown-menu {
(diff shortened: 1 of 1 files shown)


zekth:
+1

PR #634, opened 2018-11-07
(bug)preseed stats db everytime an entity created

ranjib:
To avoid bogus stats no found error before the first check interval. This causes bogus alerts in UI and confuses reef-pi users


Ranthalion:
:+1:

Issue #628, opened 2018-11-02
Allow 12-bit control of pwm for smoother dimming of lights


JohnRudolfLewis:
Allow 12-bit control of pwm so that one can control lights with much smoother dimming. With the controller that came with my LED upgrade, dimming was imperceptible. The dimming with reef-pi is very abrupt. Having the UI present things as 0-100% is ok. But when writing a value to the PWM, allow for full 12-bit control, 0-4096.

Ranthalion:
The percentage is mapped to a 12 bit value and I believe it is updated every 30 seconds.  Can you provide a sample configuration (screen shot) to show how lighting is set up?

ranjib:
Can you share your light and circuit details. The ramp up should not be very abrupt. Both auto and dirurnal profile should be fairly smooth. Current increment interval is one minute .
I do think though that we should provide both finer control of the pwm resolution as well as increment interval , it’s bit advance mode, but a good thing to have

JohnRudolfLewis:
I am using a PCA9685 that is sending its output to the LDD drivers inside one of these: http://www.stevesleds.com/Biocube-32-LED-Turbo-Upgrade-System-Plug-and-Play_p_332.html

The problem is that reef-pi uses an integer between 0 and 99 to control the duty cycle of the pwm. When scaled to 0 to 4096, the smallest increment is 40. On my LEDs, especially when just turning on, that is a very noticeable jump. The controller that came with these LEDs allowed for this finer adjustment, and the effect was much more pleasing.

Suggested fix is to take full advantage of the 12 bits. 
https://github.com/reef-pi/reef-pi/blob/master/controller/utils/pwm_pca9685.go#L48
change that int to a float with an allowed range of 0 to 1, or keep it an int and allow a range of 0-4096.

ranjib:
@JohnRudolfLewis i have added the support for specifying the % value in float instead of integer, which in turn will let us exploit the full range of 12bit controls for pca9685, and it's not specific to pwm driver. We'll update the lighting code to increment/decrement PWM values in a smaller interval than 1 minute to exploit this next.

PR #635, opened 2018-11-07
(feat)move jack specific libs to connectors package. support pwm percentage as float

ranjib:
https://github.com/reef-pi/reef-pi/issues/628


Ranthalion:
I can see how this can improve the light fading. Is there also a need to modify the UI to allow a float value for light intensity?

ranjib:
@Ranthalion not yet, I think after we merge this, we can then expose the interval in light cycle as a configurable value (its hardcodes one minute right now), and then it will be automatically taking the advantage of float %

PR #639, opened 2018-11-12
(bug)check grid values are present before accessing elements within it

ranjib:


jsx/dashboard/main.jsx
```diff
@@ -34,9 +34,15 @@ class dashboard extends React.Component {
...
       return
     }
+    if (config.grid_details === undefined) {
+      return
+    }
     var i, j
     var rows = []
     for (i = 0; i < config.row; i++) {
+      if (config.grid_details[i] === undefined) {
+        config.grid_details[i] = []
```
Ranthalion: This line is probably the source of the invariant violation.  Use continue instead.

PR #641, opened 2018-11-12
Forcing positive and scope related inputs

zekth:
Fixing input

### DIFF
--- a/jsx/dashboard/config.jsx
+++ b/jsx/dashboard/config.jsx
@@ -3,6 +3,7 @@ import Grid from './grid'
 import { connect } from 'react-redux'
 import { fetchDashboard, updateDashboard } from 'redux/actions/dashboard'
 import { isEmptyObject } from 'jquery'
+import { showError } from 'utils/alert'
 
 class config extends React.Component {
   constructor (props) {
@@ -41,7 +42,7 @@ class config extends React.Component {
     const fieldsToCheck = ['width', 'height', 'column', 'row']
     for (let prop of fieldsToCheck) {
       if (payload[prop] <= 0) {
-        console.log('invalid input')
+        showError('It seems like there is wrong data in your configuration. Please ensure you only have valid integers in fields.')
         error = true
       }
     }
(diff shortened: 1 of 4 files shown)


jsx/dashboard/config.jsx
```diff
@@ -32,29 +32,54 @@ class config extends React.Component {
...
+    const fieldsToCheck = ['width', 'height', 'column', 'row']
+    for (let prop of fieldsToCheck) {
+      if (payload[prop] <= 0) {
+        console.log('invalid input')
+        error = true
+      }
+    }
+    if (!error) {
+      this.props.updateDashboard(payload)
+      this.setState({ updated: false })
```
ranjib: With this we’ll have two different styles of validations. Is that normal ? I am ok with as long as we are all on same page and we know when to use which one . @Ranthalion comments?

jsx/dashboard/config.jsx
```diff
@@ -32,29 +32,54 @@ class config extends React.Component {
...
+        v = parseInt(ev.target.value)
+        if (!isNaN(v) && v <= Max && v >= Min) {
+          config[key] = v
+          this.setState({
+            updated: true,
+            config: config
+          })
+        }
+      } catch (ex) {
+        console.log('invalid input')
```
ranjib: showError ?? Or just drop the console statement ?

jsx/logCenter/main.jsx
```diff
@@ -27,14 +27,16 @@ class LogCenter extends React.Component {
...
-      n.push(
-        <tr className={this.getTrClass(e) + ' log-entry'} key={e.ts}>
-          <td>{this.readableDate(e.ts)}</td>
-          <td>{e.emitter}</td>
-          <td>{e.type}</td>
-          <td>{e.content}</td>
-        </tr>
-      )
+      if (typeof e !== 'undefined') {
+        n.push(
```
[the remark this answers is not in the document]
zekth: kinda the same. Just another way to test it.

jsx/dashboard/config.jsx
zekth: In fact this part is not really meant to be triggered because of the validation on the fields. If we trigger this error it means we have really messy data.

jsx/dashboard/config.jsx
zekth: I think we can just drop the console statement IMO

jsx/dashboard/config.jsx
Ranthalion: Updating to use Formik requires a pretty big change, so I'm OK with both patterns being in use.

jsx/dashboard/config.jsx
ranjib: Ok let's drop the console statement. Do we need the try catch block after that?

ranjib:
@zekth left a couple of minor nits 🚢

jsx/dashboard/config.jsx
zekth: And to be honest i don't think Formik fits in this part at all.

PR #655, opened 2018-11-29
Fix doser calibration bug

### LINKED ISSUE #654: Doser calibration does not use entered values
Doser calibration is not using the calibration values entered into the calibration request.  Instead, the doser settings are loaded and run.  

The calibration runs are also captured in the dosing charts.  I don't think it should be listed as a dosing event, but I could easily be convinced otherwise.  To prevent the calibration from showing in the chart, statsMgr will be null for calibration runs.

Ranthalion:
Use calibration settings to run doser and exclude statsMgr to suppress charting
fixes #654

### DIFF
--- a/controller/doser/pump.go
+++ b/controller/doser/pump.go
@@ -2,10 +2,11 @@ package doser
 
 import (
 	"encoding/json"
+	"log"
+
 	"github.com/reef-pi/reef-pi/controller/connectors"
 	"github.com/reef-pi/reef-pi/controller/types"
 	"gopkg.in/robfig/cron.v2"
-	"log"
 )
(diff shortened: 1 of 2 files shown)


ranjib:
@Ranthalion I was thinking of more in this line: https://github.com/reef-pi/reef-pi/pull/658 let me know what is your opinion.

Ranthalion:
> @Ranthalion I was thinking of more in this line: #658 let me know what is your opinion.

Sure. #658 is a bit more clear without the nil reference.

Ranthalion:
Closing this pull request since #658 provides a fix.

PR #662, opened 2018-12-02
Drivers/HAL: Add a drivers package

theatrus:
`drivers` is a new hardware abstraction layer for reef-pi, which
provides a pluggable interface and consolidation for the variety of
modules in the reef-pi code base. A driver is designed to provide a
named entry point, and can have a variety of interfaces representing
those capabilities. For example, Input, Output, PWM, Temperature, pH,
etc.

`drivers` is currently backend only, none of the jacks/inlets/outlets API
has changed, but they have all been converted to use the drivers package
with a fixed driver and pin name mapping.

API changes and expansion is probably best left for a second PR.


controller/drivers/rpi/pwm.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/reef-pi/reef-pi/controller/types/driver"
+
+	"github.com/reef-pi/rpi/pwm"
+)
+
+type rpiPwmChannel struct {
+	channel int
+	name    string
+	pwm     *rpiPwm
+}
```
ranjib: pwm *rpiPwm should not be a member, its convoluted . lets call this just PWMChannel, and PWMDriver

theatrus: Fixed

theatrus:
Some revisions:

- Expanded a bit of the driver API
- I hard wired Inlets to always use the "rpi" driver. In tests this uses the "mock" driver. This is obviously sub-optimal but serves to validate the basic flow until we can take another pass and expand the UI correctly. I'll do the same with outlets and jacks. I'd like to pickup the UI and data model changes in a second PR.

theatrus:
@ranjib Some more poking at this, with some options:

a) Do nothing and keep all this here. I actually don't like this.
b) Decompose `types/` into its own repo with no dependencies. This will help root the interfaces that the system expects from its various modules, especially if we are establishing a contract between the various modules to conform to an interface

I'm leaning on `b` as it will add a common type point.  I'm first going to work in #662 to cleanup the test logic and remove interface returns where I can (slice returns here are notoriously annoying)

ranjib:
I am cool with it. We can also offload the driver/interface bits to periph.io , if its usable.

theatrus:
Testify and pkg/errors are gone will work on package relocation next

theatrus:
Waiting for dependent PRs before cleaning up this one to match imports

theatrus:
note the real hardware drivers are simply commented out until 
https://github.com/reef-pi/drivers/pull/4
https://github.com/reef-pi/rpi/pull/13

land.

I left the mock drivers here.

controller/connectors/outlet.go
```diff
@@ -54,24 +23,43 @@ type Outlet struct {
 	Reverse   bool   `json:"reverse"`
 }
 
-func (o Outlet) IsValid() error {
+func (o Outlet) outputPin(drivers *drivers.Drivers) (driver.OutputPin, error) {
+	pindriver, err := drivers.Get("rpi")
+	if err != nil {
+		return nil, fmt.Errorf("inlet %s driver lookup failure: %v", o.Name, err)
```
ranjib: typo: outlet %s driver lookup failure

controller/connectors/outlet.go
```diff
@@ -54,24 +23,43 @@ type Outlet struct {
...
+
+func (o Outlet) IsValid(drivers *drivers.Drivers) error {
 	if o.Name == "" {
 		return fmt.Errorf("Outlet name can not be empty")
 	}
-	_, ok := ValidGPIOPins[o.Pin]
-	if !ok {
-		return fmt.Errorf("GPIO Pin %d is not valid", o.Pin)
+	if _, err := o.outputPin(drivers); err != nil {
+		return fmt.Errorf("outlet %s did not get associated with a driver pin", o.Name)
```
ranjib: this will confuse us, its masking the original error, better to just return the original error instead of creating new one, since this is part of the create logic caller already know name and pin of the outlet

controller/connectors/outlet.go
theatrus: fixed :+1:

PR #665, opened 2018-12-06
(chore)make pca9685 address configurable

ranjib:


theatrus:
Note to self: backport in #662

PR #671, opened 2018-12-15
(chore)Move modules to separate folder from top level controller folder

ranjib:
Moving all module specific code to separate folder. They should not be at the same level as the core controller specific things , such as connectors, drivers etc,


theatrus:
+1

Issue #667, opened 2018-12-08
Cookie expiration for API connections


sfgabe:
Xpost from reef2reef:
I've been sending commands and getting data to/from reef-pi via the API (from/to Home Assistant) for about a month now and I have to say it works flawlessly @Ranjib. Fantastic job on that!

One question though - is there a way to disable the cookie expiration? It seems like I'm getting about two weeks before I have to update the auth code by logging in and capturing a new one manually. I'm not super worried about security since it's only LAN anyway.

zekth:
It's possible, just have to set an environment option to handle the endless cookie. @ranjib your thoughts?

Here is the place where it needs to be changed:
https://github.com/reef-pi/reef-pi/blob/master/controller/reef_pi.go#L37

Documentation:
http://www.gorillatoolkit.org/pkg/sessions
https://github.com/gorilla/sessions/blob/master/options.go

sfgabe:
Hi @ranjib I see this was implemented and quickly reverted when it broke some stuff. Are there plans to get it working? If not is there a local workaround I could use? I don't feel I need a password security-wise, is there a way I can bypass auth?

zekth:
I'll check this.

PR #682, opened 2019-01-01
(chore)dont expire cookies

ranjib:
https://github.com/reef-pi/reef-pi/issues/667


zekth:
nice :)

ranjib:
btw, this broke master. instead of never expiring, this actually deletes all cookies in the session, and breaks signin process

zekth:
I'll dig on this to find a fallback instead of setting an enormous max age for the cookie.

PR #683, opened 2019-01-03
(feat)Add driver CRUD ui

ranjib:
Allow creating, deleting, updating and listing drivers through UI

TODO
- Ability to use Different configuration UI for a different type of driver
- driver selection UI across all connectors

@theatrus @zekth I think this is good to merge. By no means we are done, since there are some bits left. But the basic gears are all in. including driver CRUD api, UI, unit tests etc. What remained is custom configuration ui for individual drivers, driver selection for new connectors, and better unit test coverage across backend and frontend code
I expect another couple of PRs at least as we chalk out things. But merging this now will get master working again with driver CRUD features to develop ph board and base board support. That in turn will help us testing the new driver API/UI.
@zekth I could use some help on the frontend side of things,


theatrus:
Does _making_ drivers make sense? Having config per driver is nice, but lacking any decent loadable module support not sure creating makes as much sense?

ranjib:
one can have multiple instances of each type of drivers, like ph sensors or multiple pca9685 etc. In those cases, each of the driver may be named different, and can have different pwm frequency etc etc.
I am adding loadable module support. This is part of that work. The configs are jsin.RawMessge and will be passed on to the factories. I'll create an abstract factory that will introspect type and return the type specific factory
We'll have fixed set of types.

theatrus:
Sounds good - running drivers out of process is exciting, as long as we are comfortable with drivers being able to represent more dynamic channel/pin/etc maps (I have a few ideas here on the USB front with more hot-plug support)

PR #716, opened 2019-02-01
Populate the doser calibration fields with the doser speed and duration

kevinmic:


zekth:
This looks fine for me but make sure to run:
`npm run js-lint`
to fix the lint using standard and update your PR to make the build pass :)

Issue #713, opened 2019-01-31
Allow partial second duration for dosing pumps


kevinmic:
In trying to fine tune my dosing pump I have found that adjusting the "power" level is neither intuitive nor linear when dosing multiple times a day.  Adjusting the time however is both intuitive and linear. I would like the dosing pump feature changed so that it allows for sub second dosing.

My goal is to dose 100 ml per day. Rather than dosing all at once I prefer to dose once per hour. This means I run the pump for about 5 seconds every hour. As I try and fine tune the pump I find that adjusting the "power" level does not produce linear results. I calibrated the pump at 100% for 60 seconds produced 56 ml.  I decreased the power from 100% to 90% and it dropped amount dosed to 54 ml (4%) change.  I decreased from 90% to 80% and once again the change was not linear. Each time I am having to run a 60 second calibration in an attempt to guess the power level that will produce the results I want. If however I were able to increase/decrease the time by partial seconds I believe the changes would produce both predictable and linear results. If in the future if I find that my dosing pump is slightly under producing or over producing all I have to do is adjust the time slightly and I can expect the amount dosed will change similarly.

I have experience in programming Java and have done a little Go. I would be happy to help with the work if someone could help with the desired design and finding my way around the code.

ranjib:
This is because we are using/recommending DC motors for dosing pumps. Bulk of these issues will go away one we move to stepper motor based dosing pumps.
I think allowing sub-second duration is valuable. We can just accept float/real values in duration field (and still treat it as second) to support that.
On a side note, did you tried playing with dilution of your dosing liquid?

kevinmic:
I have a smaller 30 gallon tank and need to dose 10 ml of calcium and alkalinity per day. Due to my own pi dosing project I started diluting my solution so 9 to 1 so would dose 100 ml instead of 10.  Further dilution would force me to either use a container larger than 1 gallon or fill the container more often.  (Currently lasts for about 1 month).

For the stepper motor dosing pumps do you have a source location for those? With my own pi doser I originally tried 3d printing a peristaltic pump to work with a stepper motor but kept struggling with having enough power in the motor to turn my pump. I eventually gave up and went to a dc motor.

kevinmic:
#718

kevinmic:
This has been merged.

PR #717, opened 2019-02-01
Currently the DosingPump.regiment.duration is being stored in nanos

kevinmic:
Currently the DosingPump.regiment.duration is being stored in nanos but being interpreted in seconds. This severely limits how the duration value can be used via the api. These changes make it so DosingPump stores a true duration and the UI deals with the fact that durations returned from GO are in nano-seconds.

This is related to https://github.com/reef-pi/reef-pi/issues/713 

This is a preliminary PR to get your opinion on the change. This issue with durations is repeated throughout the code, and in my opinion should be fixed everywhere. Unfortunately changing to use duration correctly makes it a breaking changes for anyone currently using it. I added code to try and support old and new but am not sure if that was the right way to go. (I couldn't find any database update scripts so that I might apply an update script to fix existing database values). Please give feedback on what you think, what I missed, what you want changed.

### DIFF
--- a/jsx/doser/doser.test.js
+++ b/jsx/doser/doser.test.js
@@ -34,7 +34,7 @@ jest.mock('utils/confirm', () => {
 describe('Doser ui', () => {
   it('<Main />', () => {
     let mock = {
-      dosers: [{ name: 'bar', regiment: { duration: 10000000000} }]
+      dosers: [{ name: 'bar', regiment: { duration: 10000000000 } }]
     }
     const m = shallow(<Main store={mockStore(mock)} />)
       .dive()
--- a/jsx/doser/main.jsx
+++ b/jsx/doser/main.jsx
@@ -141,7 +141,9 @@ class doser extends React.Component {
 
 const mapStateToProps = state => {
   const dosers = state.dosers
-  dosers.forEach((doser) => doser.regiment.duration = secondsFromNano(doser.regiment.duration))
+  dosers.forEach((doser) => {
+    doser.regiment.duration = secondsFromNano(doser.regiment.duration)
+  })
   return {
     dosers,
     jacks: state.jacks


zekth:
For this part i think we have to do sort of a versionning of the datas to check if it's old or new format. Maybe adding another field and doing a fallback on it?

BTW, this means some changes into the UI too ?

kevinmic:
Most of my changes were to the UI. Your talk of versioning made me realize I missed a change to the UI though for backwards compatibility.

What are your thoughts if I add a "version" flag (int32) to the "pump" object. I will have the UI set "version" to 1. 
In the GO code I would check "if (pump.version <= 0) duration = duration * time.Second"
In the UI on load of data I would check "if (pump.version > 0) -- divide duration by nanoSeconds"

Thoughts?

Ranthalion:
I'm surprised the regiment duration is being used as nanoseconds.  There are other places that use a seconds duration, so I wonder if we have the same problem in those areas, as well?  If not, I'd prefer to see the doser updated to be consistent with the other behaviors.
I'd also prefer to avoid introducing a conversion to nano steps in both client and server if possible.

kevinmic:
Currently in all the areas where a duration is passed into the API they are all nano seconds. Then in the Go Code you are converting from nano to seconds/minutes by multiplying "duration * time.Second".  time.Second is "1000000000" or the number of nano seconds in a second. 

I can see your point that introducing conversion to/from nanos in the UI is a bit awkward. My main goal is to allow partial seconds, and with the duration stored the way it is this is not possible. 
An alternative way would be to change the "duration" field from a "time.Duration" to an "float64".  Then I 
could pass 5.5 from the UI. In the go code I could then convert the float64 to a time.Duration.
`
time.Duration(int64(d * float64(time.Second.Nanoseconds()))))  
`
This would also make it so nano seconds would not be exposed in the UI and the UI changes would be minimal.

Thoughts?

PR #718, opened 2019-02-02
Changing DosingPump.regiment.duration to a float64

kevinmic:
Referencing the conversation in https://github.com/reef-pi/reef-pi/pull/717

My goal is to be able to allow partial second dosing with a dosing pump (5.25 seconds). Currently the pump.regiment.duration is a "time.Duration". This field isn't really being used as a true time.Duration and as is represents a int64. I want to change it to be a float64 to allow decimal point precision.

Advantages to this way over the previous PR
1. Nanoseconds are not exposed through the API.
2. Not a breaking change (database and ui will work fine if I change it)
3. Much simpler change

A couple other thoughts. When I design an API I prefer to express the unit of a time field in its name rather than leaving it ambiguous. In this case I would change "duration" to "durationSeconds" so that any consumers of the API will know what it is without having to go search the backend code for usage. This would be a breaking change and so I am not suggesting it be changed. But for future versions of the API you might consider it.

### DIFF
--- a/controller/modules/doser/pump.go
+++ b/controller/modules/doser/pump.go
@@ -89,7 +89,7 @@ func (c *Controller) Update(id string, p Pump) error {
 }
 
 func Validate(p Pump) error {
-	if p.Regiment.Duration <= 0 {
+	if p.Regiment.Duration < 0 {
 		return errors.New("Duration must be greater than 0")
 	}
 	return nil
--- a/controller/modules/doser/pump.go
+++ b/controller/modules/doser/pump.go
@@ -2,6 +2,7 @@ package doser
 
 import (
 	"encoding/json"
+	"errors"
 	"log"
 
 	cron "gopkg.in/robfig/cron.v2"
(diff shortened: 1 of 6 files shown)


kevinmic:
I try to read why the CI build failed but I can't see the reason. Can help me read the output from the CI build so I can figure out what I need to fix.

Ranthalion:
It looks like CI has found 2 issues.  
1. In go, the DosingRegiment struct and CalibrationDetails struct are not formatted consistently and recommends running 'make imports'.  (see lines 624 -650 in CI log)
2. In JS, there are multiple style and formatting issues.   Many of these may be due to the editor you are using, but can be fixed by running "make standard". (lines 965+)

kevinmic:
Thanks for the suggestion. I was using go playground to figure it out and it seemed way too complicated.

kevinmic:
One issue I have noticed is that float64 allows for negative numbers. I am looking at how you do validation during create/update.

kevinmic:
I am done with this. Is there anything special I need to do to get it merged?

PR #735, opened 2019-02-12
(bug)Don't register and remove zombie drivers

theatrus:
Scenario:

- Create ph_board driver with invalid I2C address
  - Call to register driver fails with error, not storing the driver in the drivers slice.
    - UI shows driver instance, as it was still written to the
      underlying store.

This fixes:

- Removing a driver which returned an error on registration from the
  store (consistency)
- Allowing the removal of a driver which is in the store but not in
  the loaded map (which could happen after the driver fails to
  register due to removed hardware on reload)
- Closing the driver upon removal.

### DIFF
--- a/controller/drivers/main.go
+++ b/controller/drivers/main.go
@@ -111,6 +111,7 @@ func (d *Drivers) Create(d1 Driver) error {
 		return err
 	}
 	if err := d.register(d1, factory); err != nil {
+	    _ = d.store.Delete(DriverBucket, d1.ID)
 		return err
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


ranjib:
@theatrus let’s juts move the store.create call after register, instead of creating first and then deleting the driver if it’s failed to register . We have to remember that this will mean we won’t be able to add driver that can not be registered due to some reason

theatrus:
@ranjib Sounds good - forgot ID assignment was not required for register

theatrus:
@ranjib Actually, no go:

https://github.com/reef-pi/reef-pi/blob/master/controller/drivers/factories.go#L107

We need the ID to register, which means we need to have created the object

theatrus:
@ranjib Feel free to merge when you can :-)

PR #739, opened 2019-02-14
Multilingual

zekth:
This is the multilingual implementation for reef-pi.

All the sources files are written in csv in the files located in `assets/translations`. At the moment only english and french are written. The source of all translations keys is the English one. Other csv files can be synchronized with `npm run translations:sync`. This command will create keys in other csv files where translations are missing. For example
```
// en.csv
foo,bar
```
```
// fr.csv
foo,##
```

The language is currently selected with the prefered language of the User agent, if the language is not found the fallback is set to English. The language can be changed in the Configuration - Settings panel. To make the translation works, the rerender of the whole application is done.

At the moment translations are not finished but merging this PR will avoid a lot of pain while merging this during the development of other features.

TODO: Write a guide for translation contributors

NOTE: also changed the issue templates.


ranjib:
@zekth looks good to me, but Im novice in this. How does uses choose/select the language?

ranjib:
Scratch that. you already called it out in description. merging this

PR #752, opened 2019-02-25
Repo organization

zekth:
re-organizing all the Front-end Files in the `front-end` folder. Make it clearer to read.

Also put additionnals docs in `docs` folder.

TODO:
- Move all the go code into `back-end` folder


ranjib:
@zekth so this is juts renaming jsx folder to front-end, do you see any particular value in that? Like is this a common convention? 
Thinking out loud here, we'll need UI components for manager , where the manager will provide exactly same UI (i.e. to control any other reef-pi from its own ui) + ability to add/remove/update individual reef-pi instances. I am looking for suggestions what changes we need to make here, to make reusability easier for manager work

zekth:
It's a bit more complicated than that. At the moment i've moved and cleaned the process for the build of UI and set the PR to see if the ci passes.
renaming to Front-end is not a convention but in the future the go part will move into another back-end folder. which makes it clearer for development. Sources files in folders, builds files in another and so on. At the moment everything is mixed up and if we add the clustering part in the future it could be a pain to welcome new developers.

In your vision of the manager, is it deployed with the front-end app of reef-pi or is it another on? We haven't talked about architecture of it. 

IMO we can put the manager in the reef-pi app and enable/disable it with the status of the reef-pi if it's master or slave.

zekth:
By the way i got this strange error but haven't change anything in the part of the code.

https://travis-ci.org/reef-pi/reef-pi/builds/498618241#L652

FYI i've also added some lines in the API to get the sibling folder of service for API:
https://github.com/reef-pi/reef-pi/pull/752/files#diff-63c4aa33cafa466b59543659c003f27bR104

This make it clear where the ui has to be for any context.

zekth:
Found the issue of the strange error:
https://github.com/reef-pi/reef-pi/pull/752/files#diff-cf678a3382e0d04c979ea7f40aa50192R50

So it seems like ressources of test for ui and backend are mixed. I'd suggest to separate them because some actions can impact other side for example. I suggest this structure:
````
- back-end
  |-src
  |-test
- front-end
  |-src
  |-test
````

This give us availability to work on both sides but clearly separated. Used to do it on lots of project and working well. Your thoughts?

**Comment by theatrus** [comment]
Go doesn't split tests and source, and the path name is part of the package name so I suggest keeping `controller` or something similar here.

I'll also call out this unofficial guide which I usually use
https://github.com/golang-standards/project-layout

**Comment by zekth** [comment]
@theatrus right. But keeping test ressources in a go folder so? Because some ressources used in go test were in the ui test folder.

**Comment by ranjib** [comment]
I am ok with this. I understand where you are going, and I am treating this as transition state. I expect we'll go through some more PR like this (happy to help here, as I am working on the manager bits). @theatrus I am thinking of moving the go's test fixtures to same package under testdata, i think it will be cleaner this way.

On the manager backend /frontend code., currently i have a full blown http client that i'll use to talk to controller from manager in reef-pi/client repo and the manager code in reef-pi/manager (super WIP), but i plan to update the commnd/main.go to  hook up manager plumbing there. i.e. 
```sh
reef-pi manager # starts the manger
reef-pi controller # starts the controller
reef-pi reset-password # reset password
reef-pi export/import/client # for future use cases
```
in short, though client and manager in different repo, they all are hooked up at the command level, i expect the ui code to make little modular so that same ui can be used across controller/manager , except the redux actions will be curried with some extra stuff, and certain components will be not shown in either manager or controller. 

I am still not decided on whether to keep all the codes in separate repos and introduce gitsubmodule or dep (go dependency manager) or just pull in all the code in this repo..  
another to thing to consider is if we should move the front end to a whole different repo, and controller to another repo and keep the reef-pi repo only about build stuff, commands (execution entry points) e etc.. 
open to all suggestions.. this is a big deal, and i see pain in both ways :-)

ranjib:
btw , this is good to merge, assuming @theatrus is not terribly upset :-)

zekth:
IMO separating frontend and backend sides is not really revelant because they are tied together.

ranjib:
:shipit:

PR #756, opened 2019-03-03
(chore)Monorepo

ranjib:
Make drivers, hal, rpi all submodules

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,12 @@
+[submodule "hal"]
+	path = hal
+	url = https://github.com/reef-pi/hal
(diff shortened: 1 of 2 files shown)


ranjib:
Abandoning this in favor of introducing dep. For those who are curious, monorepo will require some more tech debt pay off. Currently hal.Metadata struct is part of interface definition which wrecks havoc in mono repo setup. Lesson learned, never have concrete types as part of interface definition

theatrus:
I wouldn't use submodules, either inline or use glide/dep and in 1.13+ use go modules

PR #767, opened 2019-04-13
(chore)Upgrade to go1.12 in CI

theatrus:
## Description

Its 2019, time to keep up with the Go releases.

## Related Issue

N/A

## Motivation and Context

Stay on currently released and bugfixed versions of Go


zekth:
Seems LGTM to me.

PR #768, opened 2019-04-13
(WIP)kasalink integration test

theatrus:
This ties a Kasalink driver (from my test branch) into Reef-Pi. There
are a number of things to fix:

- [ ] We need to add a config type for the hostname/IP in the UI -
  right now everything uses the properties of the PCA9685 type which
  has no string type.

Optional:

- [ ] It would be nice to change the load order to not rquire the rpi
  driver and load other drivers outside of requiring to use devmode.

Don't merge as is :)

https://github.com/PaulSRock/kasalink/pull/2


ranjib:
@theatrus let us know if you need help or when this is ready for review

theatrus:
@ranjib biggest holdup right now is fixing the Driver UI / model to support specifying at least host names/IP addresses. The model is specific to I2C devices (and named PCA9685 ;)) - we can either make a union type of the model for all config fields, or use different models per driver config.

ranjib:
@theatrus I am adding a bare bone tplink hs1xx driver to get my smart plugs going, we'll improve this as and when we need to. For the time being Im closing this changeset

ranjib:
https://github.com/reef-pi/drivers/pull/27

PR #777, opened 2019-05-10
spanish translation

trackerit1:
is that the translation into Spanish


ranjib:
Thank you @trackerit1 . Feel free to report back any error you may see. most of us wont know spanish :-)

PR #788, opened 2019-06-09
Add USER command to address security concern with Docker

kooma93:
I just read an article from American Express Technology about security concerns raised when Dockerized application runs as root:

https://americanexpress.io/do-not-run-dockerized-applications-as-root/

Hence, I added a USER command to avoid running as root once the image is built. I am not entirely familiar with the use of Docker in this project, but I thought this would still be helpful to avoid security vulnerabilities.


ranjib:
reef-pi does not run in docker, its only used for the development environment.

PR #793, opened 2019-06-24
Update UI for ph calibration and translations

### LINKED ISSUE #792: Support ph board calibration via UI
PH calibration is supported via api, however it is not compatible with the current UI

**Describe the solution you'd like**
Modify the calibration UI and go back end to support calibration.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

Ranthalion:
## Description

Update UI for ph calibration and translations
Sync translations

## Related Issue
fixes #792 

## Screenshots (if appropriate):
image


controller/modules/ph/probe.go
```diff
@@ -181,6 +187,34 @@ func (c *Controller) Calibrate(id string, ms []hal.Measurement) error {
...
+	}
+	if p.Enable {
+		return fmt.Errorf("Probe must be disabled from automatic polling before running calibration")
+	}
+
+	var calibration []hal.Measurement
+
+	//Append to existing calibration unless the point is the mid point.
+	//Receiving a mid point calibration resets the calibration process.
+	if point.Type != "mid" {
```
ranjib: do we need to chec the point.Type? we can just infer it from introspecting the number of calibration points? or is it purely for the step wise wizard like UX ?

front-end/src/ph/calibration_wizard.jsx
```diff
@@ -29,34 +30,53 @@ export default class CalibrationWizard extends React.Component {
...
-  calibrate (point, value) {
+  componentWillUnmount () {
+    window.clearInterval(this.timer)
+  }
+
+  readProbe () {
+    let id = this.props.probe.id
+    let that = this
+
+    fetch('/api/phprobes/' + id + '/read', {
```
ranjib: api calls like this does not need to go through the redux layer?

ranjib:
👍 looks good to me. left couple of comments.

front-end/src/ph/calibration_wizard.jsx
Ranthalion: I was wondering about that too.  I didn't put it in redux since it didn't really need to be part of application state.  I'll refactor to put it in redux and state.  It would be more consistent that way, and maybe it would be helpful if we changed it to use a rolling average over a small window.

I'll address this within the next few days.

controller/modules/ph/probe.go
Ranthalion: I'm open to other opinions on this, but here are my thoughts.

I think we need point.type for a couple of reasons.
1. If you only calibrate the mid point, then the calibration factory will use single point calibration.  Without point.type, I won't know to discard any existing calibration measurements.
2. If we support 3 point calibration in the future, I think it would be more difficult to make inferences based on existing calibration points. 

I could alternatively store the calibration measurements on the UI, then ask the user to commit them once they are happy.   That's not my preference since it changes the current expectations and could possibly result in a confusing user experience if the don't submit the measurements after collecting the points.

Ranthalion:
Moved current ph reading to redux store and refactored calibration wizard to use a more react friendly modal implementation.  I believe this is ready for another review and possibly merge.

zekth:
Seems like you're not using the same EOL everywhere, like on the translations. Are you using LF? Do you use editorconfig?

PR #798, opened 2019-07-04
add Chinese support

chendx79:
add Chinese support


ranjib:
Thank you @chendx79 , really appreciate this.

PR #799, opened 2019-07-10
Add portuguese language

ViPeR5000:
add portuguese from Portugal language.

### DIFF
--- a/front-end/assets/translations/pt-pt.csv
+++ b/front-end/assets/translations/pt-pt.csv
@@ -6,7 +6,7 @@ disabled,desativar
 cancel,cancelar
 done,Feito
 loading,Carregar
-documentation,Documentação
+documentation,DocumentaÃ§Ã£o
 errors,Erros
 edit,Editar
 analog_input,Entrada Analogica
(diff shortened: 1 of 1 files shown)


front-end/assets/translations/pt-pt.csv
```diff
@@ -0,0 +1,195 @@
+Key;Translation
+ok;Ok
+select;Selecionar
+enabled;activar
+disabled;desativar
+cancel;cancelar
+done;Feito
+loading;Carregar
+documentation;Documenta��o
```
zekth: Seems like there is some encoding problem in your file

zekth:
Thanks for the contribution, but there is some encoding issues.

front-end/assets/translations/pt-pt.csv
ViPeR5000: Ups my mistake

new version

front-end/assets/translations/pt-pt.csv
```diff
@@ -0,0 +1,195 @@
...
+alerts,Alertas
+unit,Unidades
+configure,Configurar
+language:language,Lingua
+language:fr,FranÃ§ais
+language:en,English
+language:de,Deutsch
+language:es,EspaÃ±ol
+language:it,Italiano
+language:hi,à¤¹à¤¿à¤¨à¥à¤¦à¥€
```
zekth: ew?!

ranjib:
Thank you @ViPeR5000

PR #800, opened 2019-07-12
add portugese

zekth:
As https://github.com/reef-pi/reef-pi/pull/799 has been merged. Activating it in reef-pi


ranjib:
@zekth rebase please

PR #803, opened 2019-07-14
(chore)Track translation

ranjib:
@zekth these translation files are getting updated everytime i invoke make. these are already tracked (checked in ) hence removing them from gitginore and adding the updates. Let me know if its expected to be ignored, i can do some gitfoo to untrack them and then delete them.


zekth:
At first it was a mistake on my side, because only the CSV file have to be tracked, because json files are made dynamically with `npm run translations`. So `make` run `npm run build` which run translations and then webpack. IMO those json files have not to be tracked. @Ranthalion have made some comments about it later this month in slack.

ranjib:
Ack. I can fix this i think, let me take a stab. Meanwhile I dont why travis is breaking on this one npm package, we dont have caching enabled, so this should not be due to that.

zekth:
> Ack. I can fix this i think, let me take a stab. Meanwhile I dont why travis is breaking on this one npm package, we dont have caching enabled, so this should not be due to that.

Emit the `package-lock.json` from your commit. Seems like there is a conflict with the `package.json` travis is checking out.

PR #810, opened 2019-08-07
(chore) driver ui update to support tplink devices

ranjib:
tplink drivers are wifi based hence there address configuration can be string ("ip:port"), uplike the i2c based drivers. This is a ugly hack to selectively convert the address to string or number by driver type introspection. I would prefer to offload this to yup if possible, but could not figure it out in short time.


Ranthalion:
Yup can do smarter client side validation, but it sounds like this will continue to be a kind of problem since different drivers will have different fields.  

I like the addition of classifying the types of drivers, maybe that could expand to contain the additional metadata needed to determine how to parse the parameters for that specific driver.  Or maybe the driver parameters should be modeled as Key/Value pairs.  

I think this is good enough as it is.  I expect additional friction with drivers as more functionality is added.  I think the important aspect is to consider how we can keep the database format compatible as people upgrade in the future.

Ranthalion:
What do you think about moving the driver parameters definition into the Driver and maybe the Metadata struct?

The Driver interface should have GetParameters, LoadParameters, and ValidateParameters functions.
The parameter struct could define Name, DataType, Value, and maybe Order (to support automated UI generation and validation)

Instead of passing individual parameters to the New[Driver], maybe it is parameterless and is followed by a call to LoadParameters.  

If this seems reasonable, I can work through the changes for PCA9685 driver as a proof of concept pull request.

ranjib:
@Ranthalion we can try that, but I'm not sure if it will be significantly cleaner. Since go does not have a notion of generic with strong types the Parameter (or some component of it) will likely be  interface{} type, currently config which is []byte type is doing something very similar.

Issue #830, opened 2019-09-20
Add UI support for new timer API


ranjib:
3.0 timer api has been updated to include three different types of targe
- reminder, equipment or macro
- The core scheduling fields (second, minute, hour, day) has been extended with month and day of the week. Here is an example of new timer with a equipment target
- The target field now encapsulate all time type specific information. For equipment it looks like this
```json
{
  "id": "2",
  "name": "Pump 1",
  "enable": true,
  "type": "equipment",
  "month": "*",
  "week": "?",
  "day": "*",
  "hour": "*",
  "minute": "*",
  "second": "1,31",
  "target": {
    "id": "2",
    "revert": true,
    "on": true,
    "duration": 15
  }
}
```
- For macro target 
```json
{
"id": "2",
  "name": "Pump 1",
  "enable": true,
  "type": "macrot",
  "month": "*",
  "week": "?",
  "day": "*",
  "hour": "*",
  "minute": "*",
  "second": "1,31",
   "target": {
     "id": "1"
   }
}
```
- for reminder target
- For macro target 
```json
{
"id": "2",
  "name": "Notification",
  "enable": true,
  "type": "reminder",
  "month": "*",
  "week": "?",
  "day": "*",
  "hour": "12",
  "minute": "1",
  "second": "1,",
   "target": {
     "title": "test"
      "message": "test"
   }
}
```

The UI should show drop down for choosing the specific macros from defined ones when timer type is selected as macro.

ranjib:
https://github.com/reef-pi/reef-pi/issues/795

PR #831, opened 2019-09-20
(chore)Timer ui

ranjib:
@Ranthalion @zekth I could use some help here. I have taken a stab at the UI bits, but facing some validation issue. I think i am not able pass the validation properties to nested component (target). I have detailed the API in https://github.com/reef-pi/reef-pi/issues/830

### DIFF
--- a/front-end/src/timers/target.jsx
+++ b/front-end/src/timers/target.jsx
@@ -1,22 +1,14 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import {ErrorFor, ShowError} from '../utils/validation_helper'
+import {ErrorFor, NameFor, ShowError} from '../utils/validation_helper'
 import {Field} from 'formik'
 import classNames from 'classnames'
 import i18next from 'i18next'
 import BooleanSelect from '../ui_components/boolean_select'
 
-const Target = ({
-  type,
-  target,
-  errors,
-  touched,
-  equipment,
-  macros,
-  readOnly
-}) => {
+const Target = (props) => {
   const equipmentOptions = () => {
-    return equipment.map(item => {
+    return props.equipment.map(item => {
       return (
         <option key={item.id} value={item.id}>
           {item.name}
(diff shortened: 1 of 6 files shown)


Ranthalion:
These changes look reasonable.  Passing validation to the various target types should work.  It sounds similar to the Lights validation.  I can probably pick this up next weekend if it's still an issue.

front-end/src/timers/target.jsx
```diff
@@ -0,0 +1,204 @@
...
+            <Field
+              name='name'
+              component='select'
+              disabled={readOnly}
+              className={classNames('custom-select', {
+                'is-invalid': ShowError('name', touched, errors)
+              })}
+            >
+              <option value='' className='d-none'>-- {i18next.t('select')} --</option>
+              {macroOptions()}
```
zekth: Why not putting the empty option directly into macroOptions()

ranjib: I didn't get what you mean. The macros list is coming from props

ranjib:
@Ranthalion sorry If this is not obvious, this PR is not working in its current. I am seeing validation error on client side, and not able to actually create the timer. I'll merge it none the less, as in its current form the timer ui is broken. I think I have the basic pieces here, but some yup foo is missing. I am hoping I'll pick it up as I do more work across lighting and other modules. Please take a look if you get a chance.

front-end/src/timers/target.jsx
zekth: You map it in `macroOptions`, which takes macros from props. So i'd personnaly adf the empty field in macroOptions instead of the jsx

ranjib:
@Ranthalion @zekth Im merging this. UI in master will be broken and I think thats ok for the time being, as its broken in whole lot of different places already. Take a look if you can on the validation failures. I suspect it has something to do with passing the hooks from timer to target component.

zekth:
Will do

PR #834, opened 2019-09-29
dependencies update

zekth:
Updating dependencies to be as up to date as we can.

Currently working on this, as Babel as changed their API.


zekth:
@ranjib could you try this branch? i'm suspicious about the test. I've tried it and it does not seem to fail on my dev env but i don't have revelant provisionning ATM. I think i may be due to issues with the test themselves.

ranjib:
I can reproduce the test failures.

zekth:
> I can reproduce the test failures.

could you copy the logs of the console when it fails? thanks

ranjib:
1/3rd of the tests are failing on .instance being nil.
```
 FAIL  front-end/src/timers/timers.test.js
  Timer ui
    ✕ <Main /> (9ms)
    ✓ <TimerForm /> for create (3ms)
    ✓ <TimerForm /> for edit (1ms)

  ● Timer ui › <Main />

    TypeError: Cannot read property 'toggleAddTimerDiv' of null

      47 |       .dive()
      48 |       .instance()
    > 49 |     m.toggleAddTimerDiv()
         |       ^
      50 |
      51 |     m.createTimer({name: 'test', type: 'reminder'})
      52 |     m.updateTimer({id: '1', name: 'test', type: 'equipment'})

      at Object.toggleAddTimerDiv (front-end/src/timers/timers.test.js:49:7)
```

zekth:
I rollbacked Standard upgrade and redux upgrade because those needs migration process. Also There is some warnings. I'll tackle it in another PR. Let's review @ranjib

ranjib:
awesome :-)

PR #837, opened 2019-09-30
Update standard

zekth:
Updating standard.

There is some work to do about the naming of handlers. As standard is not really configurable it's kind of a pain IMO. Eslint is more flexible.

Your thoughts?


Ranthalion:
This is a pretty big update for Standard.  I've gone through the change log and I can't really argue with the new rules.  It is a bit of a pain to rename the handlers, but I think it's worth it.  I actually like that Standard doesn't let you change the configuration, for exactly the reasons they have listed.  

I'm wrapping up the timer changes (and doser since Cron changed), so I can help with getting us up to date on Standard next.

zekth:
Ok I'll tackle the renaming a bit. If you can make a PR on my fork

ranjib:
Thank you 🚀

PR #753, opened 2019-02-27
Open Api 3 support and documentation

zekth:
As said in #727 here is the Open Api 3 support.

Spec file is currently in the process of writing.  The open api spec file is written in YAML and located in `build/openapi.yaml`

To generate the documentation we use the redoc-cli https://github.com/Rebilly/ReDoc which give us the flexibility to generate a no dependency html file to be served by the server. The generation of the doc is done by calling `npm run doc:api`

The documentation is reachable at this url `http://reef-pi-url/api-doc`

Documentation looks like this:
2019-02-27 10_41_06-redoc documentation

PR #840, opened 2019-10-02
(feat) auto api documentation

ranjib:
Add routes with documentations .Doc(req, resp)
```go
 r.HandleFunc("/api/inlets", e.create).Methods("PUT").Doc(nil, new(Inlet))
```
To generate api.txt, invoke set environment variable REEF_PI_LIST_API on startup
```
REEF_PI_LIST_API=1 make start-dev
```
and that will automatically generate
```
   PUT	/api/inlets
Input:
{
 "id": "",
 "name": "",
 "pin": 0,
 "equipment": "",
 "reverse": false,
 "driver": ""
}
```
We'll iterate from here on, by adding Doc calls to in all handler with appropirate req/resp types and then update the documentation format to markdown, or do some more fancy thing (default variable with example data?)

Replacing : https://github.com/reef-pi/reef-pi/pull/753

### DIFF
--- a/controller/daemon/api.go
+++ b/controller/daemon/api.go
@@ -1,40 +1,20 @@
 package daemon
 
 import (
-	"fmt"
 	"log"
 	"net/http"
-	"strings"
 
 	"github.com/gorilla/mux"
 
 	"os"
 
+	"github.com/reef-pi/reef-pi/controller/settings"
 	"github.com/reef-pi/reef-pi/controller/utils"
 )
 
 var DefaultCredentials = utils.Credentials{
 	User:     "reef-pi",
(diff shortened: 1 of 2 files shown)


zekth:
If you do it like this, why not exporting it as a json format using the OpenAPI standard? Then it can be consumed with any formater and any other third party.

ranjib:
@zekth we can definitely do that. does that give us free http clients? or any other benefit ?

zekth:
> @zekth we can definitely do that. does that give us free http clients? or any other benefit ?

It can be used by any platform to make documentation / http client / api gateway and so on.

PR #847, opened 2019-10-09
Move protocol and reset if contradictory

### LINKED ISSUE #625: Warn users when https is enabled and port is 80
reef-pi allow users to configure api/UI address (host and port) as well http/https. Currently this allow users to specify port 80 (http) while https is enabled. When configured like this user will see the webpage being download instead of rendering it. The following screenshot shows port 443 is used without https being enabled. 
screen shot 2018-10-28 at 8 13 02 am

As a mitigation measure, lets implement warning message in configuration UI (validation?) when user specifies port thats not 80 for http , and port 443 for https.

Ranthalion:
## Description

Moves https to the address field and removes contradictory settings on change.  Fix is as described in issue comments.

## Related Issue

Fixes #625

### DIFF
--- a/front-end/src/configuration/settings.jsx
+++ b/front-end/src/configuration/settings.jsx
@@ -32,6 +32,9 @@ class settings extends React.Component {
     this.updateHealthNotify = this.updateHealthNotify.bind(this)
     this.showHealthNotify = this.showHealthNotify.bind(this)
     this.handleSetLang = this.handleSetLang.bind(this)
+    this.handleSetAddress = this.handleSetAddress.bind(this)
+    this.handleSetProtocolHttp = this.handleSetProtocolHttp.bind(this)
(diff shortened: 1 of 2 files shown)


front-end/src/configuration/settings.jsx
```diff
@@ -70,6 +73,45 @@ class settings extends React.Component {
...
+
+  handleSetProtocolHttp () {
+    this.handleSetProtocol(false)
+  }
+
+  handleSetProtocolHttps () {
+    this.handleSetProtocol(true)
+  }
+
+  handleSetProtocol (value) {
```
zekth: Is it possible to add a test for this?

Ranthalion: Good point. I'll work on adding tests for this tab tonight.

Ranthalion: A few tests have been added for the new functionality.

ranjib:
Thank you :-)

PR #853, opened 2019-10-15
(bug)Convert equipment timers with revert duration to integer before submitting

ranjib:
currently, this is submitted as text/string causing error on the backend. @Ranthalion I was hoping the yup schema would do that for us. I have hacky solution for the time being. Advise if I should do something better.
@zekth we've got a few react upgrade warning across the board. Updating redux/react-redux..in this changeset, there are few reef-pi internal components that we can fix after 3.0.


Ranthalion:
Thanks for catching that issue with Timer.  Your solution is what I would recommend.  Yup it's only validating, so it doesn't do any marshalling.

PR #856, opened 2019-10-18
Add homeostasis to pH UI

### LINKED ISSUE #822: Implement ph based control ui
Similar to the temperature module, ph sensors can now control a set of equipment that will be turned on/off when ph is outside a range (min, max). 
Temperature and ph module now share a common homeostasis library for the control mechanism, and it can take either equipment or macro id for switching. A boolean homestasis.IsMacro property is used for detecting where the swithing target is macro or equipment. Check controlller.HomestasisConfig, temperature.TC, phprobes.Probe for detail.

Ranthalion:
Add homeostasis to pH UI + some smoke test fixes

## Related Issue

Fixes #822


ranjib:
yay :-) . thank you so much @Ranthalion

Issue #691, opened 2019-01-16
chiller temperature control is not suitable for chillers


lironalm:
chillers cannot be turned on and off like heaters as it damage the compressor.
hysteresis would be great and also time delays between operations

Liron

ranjib:
What if you dont combine them. Create a heater specific temperature controller with lower check interval and a separate chiller specific temperature controller with higher check intervals

lironalm:
check intervals will work, but not suitable for hysteresis. 
there needs to be a gap between the set point and the "turn off" point.
meaning, chiller turns on at >set point, turns off at set point -x

the temperature difference is required in order not to operate the compressor every check, for example:
set point 24
00:00 - tank temp = 24
00:03 - tank temp 24.2 -> chiller on
00:06 - tank temp 23.9 -> chiller off
00:09 - tank temp 24.1 -> chiller on

chiller is working too much. it will cycle with every check interval.
instead, it needs to be based on temperatures, not time. every cooling controller has hysteresis implemented, as it will damage compressors.

hope I described it ok... :-)

ranjib:
got it :)

ranjib:
we'll ship this in 3.0. Thank you.

PR #870, opened 2019-10-27
(feat)Add hysteresis support in homeostasis module. 

ranjib:
And incorporate the same in temperature and ph module. Add ui support in ph module. Will add the same in temperature module after @Ranthalion has wrapped up temperature ui work


ranjib:
https://github.com/reef-pi/reef-pi/issues/691

PR #871, opened 2019-10-28
Temperature Calibration

### LINKED ISSUE #324: Support providing an offset to the temperature probe
Cheap Dallas 1wire sensors (or knockoffs) from China, while they claim 0.5C precision, aren't usually all that accurate. 

There should be a spot in the UI on the temperature sensor to enter an offset for it.

Ranthalion:
## Description

Add calibration to Temperature UI.
Also upgrade Temperature to use Collapsible and CollapsiblePanel
Add read to temperature API to support observed property in calibration

## Related Issue
Fixes #324

## Screenshots (if appropriate):
image


ranjib:
Thank you :-)

PR #892, opened 2019-11-08
Update Lights UI to new API for version 3

Ranthalion:
Update Light UI to be compatible with new API.

Additional profiles are still pending, as well as manual/auto control.


ranjib:
🤗

Issue #880, opened 2019-10-31
3.0 - Telemetry input does not allow email to input


sfgabe:
Minor bug: telemetry "email to:" field does not allow for any input.

ranjib:
https://github.com/reef-pi/reef-pi/pull/893

PR #893, opened 2019-11-08
(bug) fix email to field under telemetry section

ranjib:
@Ranthalion @zekth I have no clue why this fixes the bug, Advice,

### DIFF
--- a/front-end/src/telemetry/notification.jsx
+++ b/front-end/src/telemetry/notification.jsx
@@ -27,7 +27,7 @@ export default class NotificationSettings extends React.Component {
   updateTo () {
     return function (ev) {
       const config = this.state.config
-      let recipients = ev.target.value.split(',')
+      const recipients = ev.target.value.split(',')
       config.to = recipients.map(s => s.trim())
       this.setState({
         config: config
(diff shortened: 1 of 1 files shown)


ranjib:
https://github.com/reef-pi/reef-pi/issues/880

front-end/src/telemetry/notification.jsx
```diff
@@ -27,7 +27,8 @@ export default class NotificationSettings extends React.Component {
   updateTo () {
     return function (ev) {
       const config = this.state.config
-      config.to = ev.taregt.value.split(',').map(s => s.trim())
+      let recipients = ev.target.value.split()
```
Ranthalion: The only actual change I see is removing the argument to split.  This should just return the array with a single element, although splitting on ',' should do pretty much the same.

front-end/src/telemetry/notification.jsx
```diff
@@ -27,7 +27,8 @@ export default class NotificationSettings extends React.Component {
   updateTo () {
     return function (ev) {
       const config = this.state.config
```
Ranthalion: this config is just a pointer to state.config, so the subsequent lines are directly changing a state variable.  We may need to clone the config object in order to call setState (just a theory).

PR #908, opened 2019-11-17
Fix Light UI error causing duplicate channels + time format fix

### LINKED ISSUE #907: Lights duplicating channels
**Describe the bug**
Channels are duplicated when adding a new light.  

**To Reproduce**
1. Go to Connectors and add a new jack with non sequential pins that do not start with 0 (e.g. 13,14,15).
2. Add a new light based on that Jack.
3. Channels are duplicated.

See https://www.reef2reef.com/threads/reef-pi-an-opensource-reef-tank-controller-based-on-raspberry-pi.289256/post-6718415

Ranthalion:
## Description
Fixes the issue of duplicating channels when creating a new light.  This PR also addresses the time parsing issue by requiring seconds in the UI.

## Related Issue
Fixes #907


ranjib:
Thank you

PR #924, opened 2019-11-26
Fix interval calculation for interval mode

### LINKED ISSUE #922: (bug) Lighting ui sending wrong number or interval points in interval mode
Steps to reproduce:
- Set light in auto mode. 
- Choose interval as profile type
- Set start time 07:30:00, end time: 19:30:00
- Set peak intensities for each provided interval points
- Save
- Backend throws error during PWM value calculation using profiles. It expected one more value in the argument list config.values ,

@Ranthalion  FYI

Ranthalion:
Fixed #922


ranjib:
🤗 @Ranthalion

PR #932, opened 2019-11-28
Add i18n loader

zekth:
Here is an improvement for the i18n. Now it uses a webpack loader, which means we'll have everything handled by webpack AND we have live reload on edition.

Now i just need to add the listing of incomplete translations for the CI. Mergeable for me now.


zekth:
Also i've tried to fix this wrong merge : https://github.com/reef-pi/reef-pi/commit/f1e1c36d9c8789005b8c45ca0de8797530385799#diff-6cbc6c390014d7a1c979691d65496024

@ranjib please review if it's ok

PR #976, opened 2019-12-08
Improve i18n flow

zekth:
This improves the i18n flow by adding a script to check the completion of the translations. Once all the translations will be updated and fixed we can hook github ci to create an issue once there is new entries in the master translations.

Feel free to comment.


ranjib:
👍 nice. love it

zekth:
> 👍 nice. love it

@ranjib is the report in the CI good for you? There is a detail and a resume at the end. For me it's enough ATM.

Also i can refactor the Travis to move to github CI to have everything centralized and i can also optimize the builds.

ranjib:
I like it. Fix the build and merge. I'll onboard go plumbing. Why we have to refactor travis ? I'll prefer to keep it untouched till we have github ci running smoothly for sometime

zekth:
> Fix the build and merge

So the Translations CI is exiting with 1 if there is missing translations ATM. Do we want to exit 1 in case of fatal error instead?

build/translations/csv-sync.js
```diff
@@ -1,37 +1,105 @@
...
+
+function deleteDeprecatedTranslations(slaveArray, masterArray) {
+  const masterKeys = masterArray.map(x => x.Key)
+  slaveArray = slaveArray.filter(x => masterKeys.includes(x.Key))
+  return slaveArray
+}
+
+function syncTranslationFiles(masterTranslations) {
+  files.forEach(f => {
+    let slave = fs.readFileSync(`${PATH_CSV}${f}`, 'utf8')
```
ranjib: dialect, not slave

ranjib:
Just add the i18n strings in the translation files, that should pass the build. We can have separate target to check if the individual dialects have their respective values populated or not. That does not need to be verified at CI level. It will be nice to sort keys of these file and rewrite if needed.

zekth:
> Just add the i18n strings in the translation files, that should pass the build. We can have separate target to check if the individual dialects have their respective values populated or not. That does not need to be verified at CI level. It will be nice to sort keys of these file and rewrite if needed.

I think there is a missunderstood in here. The build process is still the same and is not armed by this PR. This CI job is only checking the completion of the translations, it does not have any impact on the build itself.

build/translations/csv-sync.js
zekth: so by this, i guess `master` should be renamed as `reference` ?

ranjib:
@zekth i was talking about consolidating all CI works under github CI from travis,

zekth:
So now the CI will only break in case of missing key in Dialect from the reference language which is english. When you do the translations:sync now it will synchronize but also reorder the keys in the CSV to make everything more maintainable.

@ranjib i think it fits our needs

zekth:
Merging it @ranjib ? i don't get why the build is red. Is the ci flaky?

ranjib:
There’s a test with race condition :-/. Let me fix it. It’s annoying me for
some time, and it’s introduced by me :-(

PR #992, opened 2019-12-14
( bug)fix data race and run tests in parallel

ranjib:


danmrossi (commented):
controller/modules/ph/probe.go 

log.Println("ERROR: ph sub-system: Failed to deleted readings for probe:", id)

deleted should be changed to delete

PR #1046, opened 2020-02-09
(chore)add validation to be used before deletion

ranjib:
Add interface method to allow detecting if a given dependency is being used by any subsystem.


controller/modules/ato/controlller.go
```diff
@@ -105,3 +106,44 @@ func (c *Controller) Control(a ATO, reading int) error {
 		return sub.On(a.Pump, true)
 	}
 }
+
+func (c *Controller) InUse(depType, id string) ([]string, error) {
+	var deps []string
+	switch depType {
+	case storage.EquipmentBucket:
```
zekth: what about something like:
```go
	atos, err := c.List()
	if err != nil {
		return deps, err
	}
	for _, a := range atos {
		add := false
		switch depType {
		case storage.EquipmentBucket:
			add = a.Pump == id && !a.IsMacro
		case storage.InletBucket:
			add = a.Inlet == id
		case storage.MacroBucket:
			add = a.IsMacro && a.Pump == id
		default:
			return deps, fmt.Errorf("unknown dependency type:%s", depType)
		}
		if add == true {
			deps = append(deps, a.Name)
		}
	}
	return deps, nil
```

ranjib: @zekth whats the benefit of this route? It add another if/else and temporary variable? Just code reduction?

zekth: it's more about readability than code reduction even if it does both.

Issue #1047, opened 2020-02-09
Setup smoke testing in travis


ranjib:

ranjib:
https://github.com/reef-pi/reef-pi/issues/1071

PR #1048, opened 2020-02-09
Add smoke test

zekth:
fixes https://github.com/reef-pi/reef-pi/issues/1047

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,4 @@ script:
   - make deb
   - chmod +x ./build/smoke.sh
   - ./build/smoke.sh
+  - node front-end/test/smoke.js http://localhost:80/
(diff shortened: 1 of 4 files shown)


zekth:
@ranjib ready to ship IMO. Not the best implementation but can be tackled as we move to github actions.

ranjib:
sweet. The smoke test is broken currently :-(

zekth:
weird it passes the CI :|
maybe npm swallows the error?

ranjib:
@zekth looks like travis is passing even when the smoke test is failing.

ranjib:
@zekth smoke test needs to run in devmode otherwise reef-pi will try to bind on port 80, and in CI it does not have sufficient permission to do so. Why the bash script is failing silently is a separate issue we have to diagnose

PR #1056, opened 2020-02-26
Driver config

### LINKED ISSUE #1019: Componentize driver configuration ui
Right now reef-pi used a single text box for drivers specific configuration during the create or modify logic. This works for some cases , like ph driver has only one configuration (i2c address) but  sub-optimal for some other drivers, e.g. pca9685 have i2c address and port. 

Make driver create/modify UI componentized, as in each driver should be able to specify thei own configuration UI.

Ranthalion:
## Description
Move driver parameter into the individual drivers.  Use hal.DriverFactory and implement parameter validation.

image

## Related Issue

Fixes #1019


ranjib:
🏆 🎩 🚤 🍠 🍬

Issue #1021, opened 2020-01-07
(chore)Address react warning messages


ranjib:
These are the warning messages popping up on javascript console under development environment (for production these are disabled). Most of the warnings are from deprecated lifecycle hooks, some of which are from libraries we are using.
```
react-dom.development.js?61bb:12357 Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move code with side effects to componentDidMount, and set initial state in the constructor.
* Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: Summary
printWarning @ react-dom.development.js?61bb:12357
lowPriorityWarningWithoutStack @ react-dom.development.js?61bb:12378
ReactStrictModeWarnings.flushPendingUnsafeLifecycleWarnings @ react-dom.development.js?61bb:12542
flushRenderPhaseStrictModeWarningsInDEV @ react-dom.development.js?61bb:25689
commitRootImpl @ react-dom.development.js?61bb:24937
unstable_runWithPriority @ scheduler.development.js?1911:697
runWithPriority$2 @ react-dom.development.js?61bb:12149
commitRoot @ react-dom.development.js?61bb:24922
finishSyncRender @ react-dom.development.js?61bb:24329
performSyncWorkOnRoot @ react-dom.development.js?61bb:24307
eval @ react-dom.development.js?61bb:12199
unstable_runWithPriority @ scheduler.development.js?1911:697
runWithPriority$2 @ react-dom.development.js?61bb:12149
flushSyncCallbackQueueImpl @ react-dom.development.js?61bb:12194
flushSyncCallbackQueue @ react-dom.development.js?61bb:12182
scheduleUpdateOnFiber @ react-dom.development.js?61bb:23709
enqueueSetState @ react-dom.development.js?61bb:13994
Component.setState @ react.development.js?72d0:325
eval @ app.jsx?863a:22
Promise.then (async)
componentDidMount @ app.jsx?863a:21
commitLifeCycles @ react-dom.development.js?61bb:22101
commitLayoutEffects @ react-dom.development.js?61bb:25344
callCallback @ react-dom.development.js?61bb:336
invokeGuardedCallbackDev @ react-dom.development.js?61bb:385
[...]

PR #1145, opened 2020-05-12
(bug)Fix camera ui when no images are present

ranjib:
- Also get rid of deprecated react lifecycle method usage

### DIFF
--- a/front-end/src/camera/gallery.jsx
+++ b/front-end/src/camera/gallery.jsx
@@ -57,8 +57,9 @@ export default class Gallery extends React.Component {
   }
 
   render () {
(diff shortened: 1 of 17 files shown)


ranjib:
@Ranthalion @zekth can you review this if you get some free cycle. I have fixed some ui test without understanding exactly how they work. I think they hold state in between tests (collapsible list test)

ranjib:
https://github.com/reef-pi/reef-pi/issues/1021

Ranthalion:
I don't see any issues in this, however, I also don't experience any failed UI tests.

PR #1148, opened 2020-05-14
Add OpenAPI documentation

Ranthalion:
Add open api documentation.  This is almost entirely comment documentation using [go-swagger](https://github.com/go-swagger/go-swagger).

Use make spec to update the open api spec file.
Use make serve-spec to serve the spec locally using [redoc-cli](https://github.com/Redocly/redoc/blob/master/cli/README.md).

swagger.json  can be served from github to support api documentation on reef-pi.com (or even on github itself) once this PR is reviewed and merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,10 +118,6 @@ race-test:
 spec:
 	swagger generate spec /w ./commands/ -i swagger.yml -o swagger.json -m
 
-.PHONY: serve-swagger
-serve-swagger:
-	swagger serve /flavor:swagger .\swagger.json
-
-.PHONY: serve-doc
-serve-doc:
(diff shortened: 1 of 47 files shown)


zekth:
Really ncie work, just why is there `swagger.yml` and `swagger.json` as siblings? Those are different as i see.

Ranthalion:
> Really ncie work, just why is there `swagger.yml` and `swagger.json` as siblings? Those are different as i see.

swagger.yml is an input to creating swagger.json.  Go-swagger cannot generate the schema for maps if the key isn't a string, so I manually created the light definition in swagger.yml and it gets used in swagger.json.

PR #1154, opened 2020-05-26
(bug) fix panic scenarios

ranjib:
- lunar profile computation panic when full moon is set to future date
- driver manager panic during un-initialized driver registration

### DIFF
--- a/controller/pwm_profile/lunar.go
+++ b/controller/pwm_profile/lunar.go
@@ -3,6 +3,7 @@ package pwm_profile
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"math"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


controller/pwm_profile/lunar.go
```diff
@@ -32,20 +32,24 @@ func Lunar(conf json.RawMessage, min, max float64) (*lunar, error) {
...
+	}
 	return &l, nil
 }
 
 func (l *lunar) Get(t time.Time) float64 {
 	d := t.Format(_lunarDailyFormat)
 	if l.dailyProfile == nil || (l.previousDay != d && l.IsOutside(t)) {
-		l.buildDailyProfile(t)
+		if err := l.buildDailyProfile(t); err != nil {
+			return 0
```
Ranthalion, line 51: I would prefer to log something so we can troubleshoot lunar cycle not working.

ranjib: Ack. I'll add logging statement

PR #1208, opened 2020-08-03
Use nvm latest version for golang ci

### LINKED ISSUE #1207: CI Fails due to node version
CI is currently failing due to the node version.  Travis is using version 8, but we need version 11+.  

I tried splitting the CI into golang and node parts, but make deb relies on both.

Ranthalion:
Use nvm latest version for golang ci
Split ci between back end and front end
switch to npm ci to improve ci speed

Fixes #1207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,30 @@
-language: go
-go:
-  - "1.13"
-before_install:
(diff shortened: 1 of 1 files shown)


ranjib:
🙇

Issue #1087, opened 2020-03-28
Temperature calibration resetting after 2 minutes 


gminton86:
When I calibrate my temp probe it resets back to what it was before calibration. If its reading 74 but actually supposed to be 77 I hit calibrate make it 77 and it will change the temp to the right temp then a few minutes later it resets. Here is a picture of it happening.
![ptemp  calibration resetting](https://user-images.githubusercontent.com/22088453/77819004-d85f5180-70ad-11ea-8fbc-f6df41b335b9.JPG)

ranjib:
https://github.com/reef-pi/reef-pi/issues/1053

gminton86:
I dont know why but now it has fixed itself and the calibration has finally stuck

Ranthalion:
Fixed by #1211

Issue #1053, opened 2020-02-21
Temperature Calibration Lost After Edit


clownfish44:
I was editing the temp to turn on notifications and it lost the calibration. 
![image](https://user-images.githubusercontent.com/15145228/74998132-ecb48c80-5425-11ea-9381-8b64654d9783.png)

v3.2

ranjib:
https://github.com/reef-pi/reef-pi/issues/1013

Ranthalion:
Fixed by #1211

Issue #1013, opened 2020-01-02
When I try to calibrate temp sensor nothing happens what are the steps to calibrate the temp sensor in 3.0


gminton86:
sw

gminton86:
Sorry I forgot to reboot after calibrating
For the future you have to 
1.) Disable the temp sensor.
2.) Calibrate to the desired value you want.
3.) Save and then enable the temp sensor again.
4.) Reboot the system.

gminton86:
actually im still having an issue with this ranjib can you help me? After a reboot my calibration went away for my temp sensor and I still cant get it right its 2 degrees off and in the calibration I make it +2 and it changes nothing?

ranjib:
Sorry for the delay @gminton86 .Can you share some details like a screen shot  and sample logs showing the temperature values from reef-pi

gminton86:
![Capture3](https://user-images.githubusercontent.com/22088453/74092579-123fae80-4a94-11ea-85b9-6abd9894a619.JPG)
![Capture](https://user-images.githubusercontent.com/22088453/74092580-12d84500-4a94-11ea-9c06-0fef5d7f802c.JPG)
![Capture2](https://user-images.githubusercontent.com/22088453/74092581-12d84500-4a94-11ea-8352-570fca68b9ca.JPG)

gminton86:
i disable the temp sensor 
hit calibrate
my temp sensor is off by 2 degrees so i enter 2
hit apply and then reenabe the temp sensor and it stay the same 
if i enter 77.31 into the set temperature box I get an error

gminton86:
can anybody help me

ranjib:
https://github.com/reef-pi/reef-pi/issues/1087

Ranthalion:
Fixed by #1211

PR #1211, opened 2020-08-06
Temp calibration

### LINKED ISSUE #1191: After I calibrate the temp sensor I cannot change the check frequency or it will reset my calibration.
**Describe the bug**
If I change the check frequency of the temp sensor after calibrating. It will lose the calibration.

Reef Pi 3.3.1

Ranthalion:
Remove calibration from TC and created dedicated bucket.  Modify temperature calibration to be applied to the temperature sensor instead of the temperature controller.

Fixes #1191, #1087, #1053, #1013


ranjib:
@Ranthalion I'm ok with migrating the calibration detail to sensor instead of the controller. This implementation requires a db read for every sensor read. I'll prefer to cache that and read only on load/update.

Ranthalion:
> @Ranthalion I'm ok with migrating the calibration detail to sensor instead of the controller. This implementation requires a db read for every sensor read. I'll prefer to cache that and read only on load/update.

That makes sense.  The pH code does the same thing, so I'll address that behavior in both modules.

Issue #1213, opened 2020-08-08
Ability in the macro tab to wait for a sensor to be triggered


chrispo1:
It would be nice if we could wait on a sensor to be triggered with a possible timeout in a macro.

This would mainly be for automatic water changes.

turn off ATO
start drain pump
wait for sensor 1
start fill pimp 
wait for sensor 2
turn on ATO

gminton86:
Yep I like this also

ranjib:
https://github.com/reef-pi/reef-pi/pull/1215
Exposing direct access to sensor in macros will be significant changes and may lead to bugs/inconsistency, since its a higher-level abstraction (on top of controllers).  I am proposing Oneshot controllers, akin to systems one-shot unit files that will be useful in automating the above-mentioned scenarios, like filling/draining tank using a level sensor,

Ranthalion:
Adding as a branstorming idea, or more of a long term architecture change - 
We should decouple/split the producers and consumers of the sensors.  The controllers could start a routine that monitors the sensor, but is not responsible for acting on the value.   Instead, it only emits the new reading. Any interested consumers can take whatever action they want as new values appear in that bus.

PR #1215, opened 2020-08-09
One shot controllers

ranjib:
- Controller (ato,  temperature and ph) can be declared as one shot type (boolean)
- When declared oneshot controllers will not monitor sensor perpetually, instead they'll run till sensor attains the desired state. For ato its 1/true, for ph/temperature this will be within min/max range.
- "On" method for subsystems will treat oneshot controller differently and block/wait till controller run is finished.
- Macros use "On" method for executing individual steps, thus when used with macros, oneshot controllers can be used to automate steps drain or fill tank or start water change process after new mix reaches a certain temperature etc
🚧


Ranthalion:
I see where this is going, and I agree that it could work.  On the other hand, I have concerns that the entire controller is started and running.  It may not matter, but it seems it could also have unintended side effects if the ATO is also used for non-oneshot behvior.  

I think it also muddies where the elements of the ATO are stored.  Is the one shot field a property of the macro step, or is it a propery of the ATO in storage?

I commented on issue #1213, but essentially decoupling sensor reading from actions with something like a pub sub pattern could be a better alternative.

Ranthalion:
Added PR #1219 as POC for alternate idea.

chrispo1:
Can this be tested in your branch if I compile it?

ranjib:
yes but you also have to use the api to set the onshot attribute to true. I'll start working on the ui support soon

chrispo1:
Is the updated ui in version 3.5?

PR #1272, opened 2020-11-11
(feat)Database import/export

ranjib:
- Add API to download and upload db file
- Add UI element to download/upload db file
- Add restore-db command to do out of band db file replacement


package.json
```diff
@@ -45,10 +45,9 @@
     "lodash.omit": "^4.5.0",
     "mini-css-extract-plugin": "1.2.1",
     "nightmare": "^3.0.1",
-    "node-sass": "5.0.0",
     "popper": "^1.0.1",
     "popper.js": "^1.16.1",
-    "postcss-loader": "4.0.4",
+    "postcss-loader": "^3.0.0",
```
zekth, line 50: why the downgrade?

ranjib:
@zekth merging this for the time being. we'll definitely revisit the dependency issue. The UI elements for db download/import is super wonky as well, could use some love.

PR #1294, opened 2020-11-25
Rename TwoDecimal method and move to utils package

hectorespert:
## Description

Rename `TwoDecimal` method to `RoundToTwoDecimal` and move from telemetry package to utils package

## Related Issue

None

## Motivation and Context

`TwoDecimal` method is used outside the telemetry controller in other parts and uses the same test file that the `Mailer`, but the main motivation for my side is to start to known how reef-pi project works.

### DIFF
--- a/controller/homestatsis.go
+++ b/controller/homestatsis.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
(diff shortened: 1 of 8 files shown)


ranjib:
Thank you

PR #1296, opened 2020-11-28
Support translations for notification texts

hectorespert:
Add keys to support translations in the email configuration page.


ranjib:
Thank you

PR #1317, opened 2020-12-24
add systemd watchdog support

sabbene:
While submitting a PR, please make sure the build is ok with Travis CI. If not fix it, and if you need help feel free to ask :blush: .

## Description

Adding support for systemd watchdog, so reef-pi service can be restarted if it hangs.  My golang isn't great, so please let me know if there are any issues.  The current health check is just to make sure that the address that is configured for reef-pi to listen on is accessible and acking to syn packets.

## Related Issue

NA

## Motivation and Context

Add support for systemd watchdog to allow systemd to restart reef-pi if it stops responding on the port/address it is listening on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,19 @@
 # Use debian as it matches a Raspbian environment
-FROM debian:stretch
+FROM debian:buster
 
 LABEL maintainer="code@reef-pi.com"
 
 RUN apt-get update -y && \
-    apt-get install curl build-essential git mercurial -y
+    apt-get install curl build-essential git mercurial ruby-dev build-essential npm systemd -y && \
+    gem install bundler
 
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - && \
     apt-get install -y nodejs && \
     npm install -g npm
-RUN curl https://dl.google.com/go/go1.12.4.linux-amd64.tar.gz > /tmp/go1.12.4.linux-amd64.tar.gz && \
-    tar xvzf /tmp/go1.12.4.linux-amd64.tar.gz -C /usr/local
+
+RUN apt-get clean && \
+    curl https://dl.google.com/go/go1.14.2.linux-amd64.tar.gz > /tmp/go1.14.2.linux-amd64.tar.gz && \
+    tar xvzf /tmp/go1.14.2.linux-amd64.tar.gz -C /usr/local
 
 
 ENV GOPATH=/gopath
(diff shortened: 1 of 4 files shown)


ranjib:
@sabbene Thank you for the PR
It will be awesome if you can isolate the Dockerfile changes and submit as separate PR, it will help others, and run on deprecated os right now. 

I think we should keep the main controller as it is, and run the watchdog go-routine as part of the health module.  We should make it optional (i.e. user should be able to opt-in /out watchdog functionality).  Let's not change the default unit file right now (since it's opt in).
Move the systemd specific logic on controller/utils/systemd (StartSDNotify()), and remove the tcp check. A simple tcp syn check does not provide much. The controller can run even if API (bad IP configuration) fails. Also, I am not sure how the tcp check will react to bad IP or 0.0.0.0. With all this in mind, i think the tcp check does not add much value than vanilla sd notify at a periodic intervals.

ranjib (changes requested):
- Move systemd specific usage to controller/utils/systemd.go , as StartSDNotify()
- Make watchdog feature optional and under health subsystem
- Remove tcp based check.

sabbene:
Sure, I can make those changes.  Making the changes to the UI to make the watchdog stuff optional may take me a bit of time.  Don't know much react.  There probably needs to be some sort of popup if you enable the watchdog explaining how to modify the unit files, and restart the daemon?  But it does seem not super useful if there isn't some sort of health check, maybe there is some other health check that could be performed?.  I tested with 0.0.0.0 and the check is performed against localhost, there is a screenshot of tcpdump below.

image

hectorespert:
@sabbene Is this watchdog feature only for the reef-pi systemd service? Could it support [this watchdog daemon](https://github.com/reef-pi/reef-pi.github.io/pull/15/files)?

sabbene:
@hectorespert -- Is is a slightly different watchdog mechanism.  The watchdog in your link is for detecting system issues that require a reboot.  The systemd watchdog is for system daemons to check in with systemd regularly, and be restarted if there are any issues.  I supposed that the test/repair scripts for the system watchdog could be used to implement a similar feature, but that seems like a kludge.

ranjib:
@hectorespert the userland watchdog daemon predates systemd and is not required once systemd watchdog is in place for our use case at least

hectorespert:
> @hectorespert the userland watchdog daemon predates systemd and is not required once systemd watchdog is in place for our use case at least

Can Systemd watchdog protect from a block generated from other program or other block condition?

ranjib:
@sabbene checking in, still interested in this?

ranjib:
Closing due to inactivity. Please reopen if you want to revive this

PR #1340, opened 2021-01-07
Allow temperature sensors based on the MCP9808 (jc42 kernel driver) to work.

MelanieT:
This path extends the code to support temperature sensors based on the MCP9808, a digital high precision i2c temperature sensor that requires no calibration.
To use, connect a MCP9808 based sensor to the i2c bus and add i2c-dev and jc42 to /etc/modules.

### DIFF
--- a/controller/modules/temperature/api.go
+++ b/controller/modules/temperature/api.go
@@ -1,13 +1,14 @@
 package temperature
 
 import (
-	"net/http"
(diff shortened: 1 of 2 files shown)


MelanieT:
Pipeline build failed with master where it succeeded with 4.0. Fixing for master before resubmission.

PR #1341, opened 2021-01-07
Allow use of MCP9808 sensors (driver jc42)

MelanieT:
Changes the temperature controller to be able to also use MCP9808 sensors. These are high precision, calibration free temperature sensors connected via I2c.
To use, insert i2c-dev and jc42 into /etc/modules to make the new devices discoverable.
Note: This patch changes the format devices are stored in the database to allow differentiating between the module types. The sensor type for existing temperatore controllers needs to be reselected in the UI once this patch is applied.

### DIFF
--- a/controller/modules/temperature/api.go
+++ b/controller/modules/temperature/api.go
@@ -1,13 +1,14 @@
 package temperature
 
 import (
-	"net/http"
-	"path/filepath"
+    "net/http"
+    "path/filepath"
+    "strings"
 
-	"github.com/gorilla/mux"
+    "github.com/gorilla/mux"
 
-	"github.com/reef-pi/hal"
-	"github.com/reef-pi/reef-pi/controller/utils"
+    "github.com/reef-pi/hal"
+    "github.com/reef-pi/reef-pi/controller/utils"
 )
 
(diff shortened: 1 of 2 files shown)


ranjib:
@MelanieT I'll prefer to keep the current temperature controller as it is and implement mcp9808 driver like all other HAL based drivers (in reef-pi/drivers).  I intend to consolidate the temperature module into current ph module, and rename it as common sensor module. Currently they use pretty much same code except the hardcoded ds18b20 specific reading logic, that i plan to port in a separate driver.

For a new [HAL](https://github.com/reef-pi/hal) based mcp9808 driver, checkout the ph board [implementation](https://github.com/reef-pi/drivers/tree/master/ph_board) 

Let me know if you need any help

MelanieT:
Hi,

I'll have to see when I get to it, but the pointer to a similar
driver in HAL architecture was just the bit I needed.

Thanks,

Melanie

ranjib:
@MelanieT Closing this one, would be awesome if you can create the driver PR against reef-pi/drivers repo. I can pull in the support in the controller soon after that.
thank you

PR #1360, opened 2021-01-21
Sanitize prometheus metric names to a more strict filter

theatrus:
Prometheus really only wants [a-zA-Z0-9_], previously we only removed
"-" but other characters would then proceeed to blow up if you use
"/", or any number of other characters. This restricts the subset to the
smaller set of allowed characters.


zekth (approved):

ranjib:
thanks @theatrus

PR #1374, opened 2021-02-01
Correct behaviour of timer invoking a macro

chrisleavoy:
**Previously**, a timer invoking a macro would execute the macro in "reverse" and be done.  It would not revert to the previous state.

**With this change**, a timer executing a macro will execute the steps normally, and then a second time, in reverse order to revert to the previous state.

### DIFF
--- a/controller/modules/timer/macro.go
+++ b/controller/modules/timer/macro.go
@@ -16,7 +16,10 @@ type MacroRunner struct {
 }
 
 func (m *MacroRunner) Run() {
+	if err := m.c.On(m.target, false); err != nil {
+		log.Println("ERROR: timer sub-system, Failed to complete macro. Error:", err)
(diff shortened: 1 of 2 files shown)


ranjib:
@chrisleavoy i am thinking of consolidating the macro and equipment runner into one, and extend it to all other subsystem. that https://github.com/reef-pi/reef-pi/pull/1375 . This will not only solve the timer based macro reverse behavior, but will also allow ato, doser etc to be directly assigned as timer target. Currently this is done in more complex way, by creating a proxy macro and then assign that macro to a timer.

ranjib:
Merged #1375 . Timer module has the macro revert fix , and all other subsystem now integrated with a common implementation. Thank you for cathing this @chrisleavoy

PR #1390, opened 2021-02-05
More UI change suggestions

prsws:
Hi:

This version adds an optional "blank" panel (called "Unassigned") to the dashboard to use as filler, instead of the Health widget. This should improve the look and feel of the dashboard,

It fixes the missing title for the Equipment Switch Panel widget.

It also sorts the Journal entries alpanumerically by name instead of by id.


prsws:
Hi:

I'm unsure how to solve the problems in the failing test jets/builds because I haven't touched the *.test.js files. Please help.

Thanks,

José

ranjib:
@prsws its javascript lint violations fro standard: https://github.com/reef-pi/reef-pi/pull/1390/checks?check_run_id=1841695132#step:6:57

you can recreate them by `npm run standard`. There are bunch of warning and a few errors, you just have to fix the errors
```
npm run standard  | grep -v ignore
```

prsws:
> @prsws its javascript lint violations fro standard: https://github.com/reef-pi/reef-pi/pull/1390/checks?check_run_id=1841695132#step:6:57
> 
> you can recreate them by `npm run standard`. There are bunch of warning and a few errors, you just have to fix the errors
> 
> ```
> npm run standard  | grep -v ignore
> ```

Got it, thanks Ranjib!

front-end/src/temperature/readings_chart.jsx
```diff
@@ -41,9 +41,9 @@ class chart extends React.Component {
                 <stop offset='95%' stopColor={c.color} stopOpacity={0} />
               </linearGradient>
             </defs>
-            <YAxis dataKey='value' domain={[c.ymin, c.ymax]} />
+            <YAxis dataKey='value' allowDecimals='false' domain={[c.ymin, c.ymax]} />
             <XAxis dataKey='time' />
-            <Tooltip formatter={(value, name) => [value, unit]} />
+            <Tooltip formatter={(value, name) => [parseFloat(value).toFixed(2), unit]} />
```
ranjib, line 46: toFixed will cause error if parseFloat returns NaN. Either omit this or add a helper method  for two decimal precision instead of directly calling toFixed that can deal with parseFloat failures

front-end/src/drivers/main.jsx
```diff
@@ -29,7 +29,11 @@ class drivers extends React.Component {
 
   list () {
     const items = []
-    this.props.drivers.sort((a, b) => { return parseInt(a.id) < parseInt(b.id) }).forEach((d, n) => {
+    this.props.drivers.sort((a, b) => {
+      return a.name.localeCompare(b.name,
```
ranjib, line 33: it will be nice to move this logic in a separate helper method
```js
localeCompare(str1,str2) {
    return str1.localeCompare(
      str2 ,
      navigator.languages[0] || navigator.language, 
      { numeric: true, ignorePunctuation: true }
  )
}

```

front-end/src/dashboard/main.jsx
```diff
@@ -80,6 +81,15 @@ class dashboard extends React.Component {
               </div>
             )
             break
+          case 'blank_panel':
+            columns.push(
+              <div className='col' key={'chart-' + i + '-' + j}>
+                <ErrorBoundary>
+                  <BlankPanel width={config.width} height={config.height} />
```
ranjib, line 88: Im curious, do we need a responsive rechart container here? what if its just skip it and push only the parent div elemenr

ranjib (commented):
@prsws overall looks good to me. Left some comment around validation. I had to remove the toFixed calls due to errors after the previous patch, hence the extra validation request.

front-end/src/temperature/readings_chart.jsx
prsws: Added utility function two_decimal_parse.js to handle parseFloat failures.

front-end/src/drivers/main.jsx
prsws: A number of attempts were unsuccessful after a few hours trying:

.../utils/sort_by_name.js
------------------------
export const SortByName = (a, b) => {
  return a.name.localeCompare(b.name,
    navigator.languages[0] || navigator.language,
    { numeric: true, ignorePunctuation: true })
}

.../ato/main.jsx
import SortByName from '../utils/sort_by_name'
   ...
    return this.props.atos.sort((a, b) => function () {
      SortByName(a, b)
    }).map(probe => {...

For some reason it would not perform the localCompare successfully.

Turns out that the sort function is just a few characters more than the original sort by id so it's not a big setback. Left it as:

return a.name.localeCompare(b.name, navigator.languages[0] || navigator.language, { numeric: true, ignorePunctuation: true })

Thanks.

front-end/src/dashboard/main.jsx
prsws: It's not strictly necessary but in my testing I found out that the Recharts responsive container provided a cleaner, better looking UI when using cellphones and tablets.

prsws:
Doing more checks, will submit PR again soon. Thank you.

front-end/src/drivers/main.jsx
prsws: Resolved issue; added utility function SortByName.

prsws:
Completed changes requested by R. Dey.

ranjib:
Thank you

PR #1411, opened 2021-04-03
Added StayOffOnBoot checkbox to Equipments + dosing calibration modal fix

prsws:
This PR includes:

* added a checkbox to the list of Equipments to select whether they should stay off upon boot. It has happened to me that a timer or macro controlled equipment is turned on when a power outage happens. When power comes back on the equipment restarts but the timer/macro is not running anymore. This can result in, for example, a dosing pump remaining turned on for hours. If the feature works well I will add it to the ATO and Dosing capabilities.

* fixed a problem with the Dosing calibration modal that was introduced upstream after the official 4.1 release. Since I base my work on the latest master branch from upstream, the bug sneaked into my code. If you encounter problems with it please let me know.


controller/modules/equipment/controller.go
```diff
@@ -24,6 +25,16 @@ func New(c controller.Controller) *Controller {
 }
 
 func (c *Controller) Setup() error {
+	eqs, err := c.List()
+	if err != nil {
+		log.Println("ERROR: equipment subsystem: failed to list equipment outlets during Setup. Error:", err)
+	}
+	for _, eq := range eqs {
+		if eq.StayOffOnBoot && eq.On {
```
ranjib, line 33: This check  should be moved in start method , setup is used in the very begining, List can fail if the storage bucket is not setup.
```go
for _, eq := range eqs {
if eq.StayOffOnBoot { 
eq.On = false
}
c.Update(eq.ID, eq)
}

```

controller/modules/equipment/api.go
```diff
@@ -119,7 +119,8 @@ func (e *Controller) LoadAPI(r *mux.Router) {
 
 //swagger:model equipmentAction
 type EquipmentAction struct {
-	On bool `json:"on"`
+	On            bool `json:"on"`
+	StayOffOnBoot bool `json:"stay_off_on_boot"`
```
ranjib, line 123: Not sure why this is required. This attribute is already present on equipment, why we need it again in EquipmentAction

go.sum
```diff
@@ -54,6 +54,8 @@ github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25Kn
 github.com/eapache/go-resiliency v1.1.0/go.mod h1:kFI+JgMyC7bLPUVY133qvEBtVayf5mFgVsvEsIPBvNs=
 github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21/go.mod h1:+020luEh2TKB4/GOp8oxxtq0Daoen/Cii55CzbTV6DU=
 github.com/eapache/queue v1.1.0/go.mod h1:6eCeP0CKFpHLu8blIFXhExK/dRa7WDZfr6jVFPTqq+I=
+github.com/eclipse/paho.mqtt.golang v1.3.1 h1:6F5FYb1hxVSZS+p0ji5xBQamc5ltOolTYRy5R15uVmI=
+github.com/eclipse/paho.mqtt.golang v1.3.1/go.mod h1:eTzb4gxwwyWpqBUHGQZ4ABAV7+Jgm1PklsYT/eo8Hcc=
```
ranjib, line 58: rebase from master

controller/modules/equipment/controller.go
prsws: Moved to Start method and simplified StayOffOnBoot logic as requested by R. Dey.

controller/modules/equipment/api.go
prsws: Removed property StayOffOnBoot as requested by R. Dey.

go.sum
prsws: Performed rebase from master but Git reported "already up to date".

front-end/src/doser/main.jsx
```diff
@@ -24,10 +22,6 @@ class doser extends React.Component {
     this.calibrateDoser = this.calibrateDoser.bind(this)
   }
 
-  dismissModal () {
```
ranjib, line 27: CalibratonModal closes properly without this? I think we have a bug that I have to fix on master,

ranjib:
This looks awesome and ready to :ship:  @prsws  . Let me know your thoughts on the calibrationModal dismissal bit.

front-end/src/doser/main.jsx
prsws: Yes. Moreover I would suggest that the "Run" button change to "Stop" after being activated and that it actually stops the doser. The dialog ignores the user's intention to stop the calibration procedure at will and continues until the end of the entered period. Pressing "Done" only dismisses the dialog but not the dosing so this should be fixed as well. For a regular person, "Done" means **Done** so perhaps a better term is warranted.

PR #1496, opened 2021-07-15
Syncronize all node version in GitHub actions

hectorespert:
Use Node version 14 in all of  the GitHub actions.


zekth:
We'd rather read: https://github.com/reef-pi/reef-pi/blob/master/.nvmrc using some patterns like:

https://github.com/keycap-archivist/website/blob/5c5eae02f875ffc468131ad206f29d7b76caff99/.github/workflows/build.yml#L21-L31

PR #1497, opened 2021-07-15
Add Tasmota driver

hectorespert:
Add Tasmota driver factory in the factories register


hectorespert:
Pending to solve this issue: https://github.com/reef-pi/drivers/pull/82

ranjib:
🤗

Issue #1500, opened 2021-07-29
Show heater/chiller usage as percentage of time instead of toal seconds


schwabix-1311:
Currently heater/chiller usage is shown as runtime in seconds per time unit (usually 3600, or can this change somehow?).
It would be more intuitive to show a percentage instead. This is applicable to all charts showing usage info.

As an example if current chart shows 173, I tend to divide this by 3600 to see how much of available time this unit was in use.
With a percentage display you could even allow zooming in and out; the percent values would not change, while seconds would change considerably. Since the y axis maximum automatically adjusts to the highest bar it is no help to know what 100% usage would be.

Does this make sense? A new PR (separated out of my pending #1495) will follow soon to demonstrate this for heater/chiller legend, since I did not succeed yet to change the data points themselves. (The interaction between go and JS is still a white spot on my map).

PR #1502, opened 2021-07-29
Usage charts show utilization as % of time

schwabix-1311:
A sample for the proposal in issue #1500 
So far only the tooltip for heater/chiller usage has been modified.


ranjib:
@schwabix-1311 i see now.. i dont know if folks will like it or not. I think its a good move. Let me ask and take some feedback from users. We have to updated the documentation accordingly. Meanwhile, can you remove the go.mod/go.sum.orig files. they seem to be unrelated.

ranjib:
@schwabix-1311 pull and fix the tests?

PR #1503, opened 2021-07-30
Fixes to allow full translation of UI

schwabix-1311:
(hopefully) all string literals are now mapped through i18next to string tables (xx.csv)

The messages of validation (*_schema) are not yet included.

This replaces my former PR 1495, and will be followed by a 2nd PR that contains the German translation.


ranjib:
Thank you :-)

PR #1530, opened 2021-08-18
Revert "(bug)switch off ato target equipment while disabling by alert"

ranjib:
Reverts reef-pi/reef-pi#1526


ranjib:
equipment are always switched off when ato is disabled, thus this logic is redundant, we should track and fix any bug otherwise

PR #1531, opened 2021-08-18
Fix: labels ph:upper_function and ph:lower_function swapped in German UI

schwabix-1311:
mis-interpretation:  is "lower_function" the function to use when lower threshold is reached, or to reduce the controlled measurement?
Wasn't able to test the pH subsystem at time of translation. Test today showed German translation needs a swap of the two labels.


ranjib:
🙇

PR #1534, opened 2021-08-28
(chore)Docker github action

ranjib:


ranjib:
docker is getting proprietory, and im planning to just remove dockerfile altogether from the repo

PR #1535, opened 2021-08-30
Update and clean dependencies and build config

Wraul:
As promised, here is a PR with some initial work on the build config and dependency updates.

There is still things to do, but in order for the PR not to get too large I felt like this was a good point to stop for now.

~~Ohh, I almost forgot. I have increased which files are listed using standard. So they will fail when running in CI.
I was planing to fix those in a separate PR.~~ Never mind, I fixed the lint errors, so the CI should pass now.


Wraul:
I have corrected all the lint errors. So the CI should hopefully pass now.

ranjib:
@Wraul please rebase. these are all formatting or const/let declaration changes that cases lint violations? or just javascript standard formatting? 
it will be lot easier to review if we segregate the yarn task vs lint fixes vs package updates. for some of these (like package.json based direct UI dependency changes) we'll need more testing (on devices) , thus having them as independent changes will be easier for review. I'll suggest close this one and cherry-pick those commits and raise individual PRs for them.

Wraul:
@ranjib I can absolutely spit this into smaller PRs. I've just created the first one, and the rest will follow.

The thing that causes the lint CI stage to fail was a combination of multiple things. Mainly the upgraded lint tools and the fact that I now also include the test files in these checks. Though everything passes as of this PR.

PR #1555, opened 2021-09-13
Update webpack to 5.52.1

Wraul:
This also updated related packages and plugins.

Some changes was required to the webpack configuration for things to work with the new version.


ranjib:
thank you.

PR #1556, opened 2021-09-13
Update Jest to 27.2.0

Wraul:
This updates Jest to the latest version.

It was also necessary to adjust the configuration to stop the new version from complaining about it running in the wrong environment (not jsdom).

I also took this opportunity to simplify and generalize the testing related configuration and scripts.


ranjib:
@Wraul are you seeing any issue with the existing versions? Trying to understand what issue this solves.

Wraul:
> @Wraul are you seeing any issue with the existing versions? Trying to understand what issue this solves, or its just a version update

It is mostly just a version update. Though my motivation for updating dependencies is due to the following:

1. I'm experimenting with building and running Reef-Pi on NixOS and I want to avoid having to troubleshoot issues that might already have been fixed in more recent versions. 
2. I find it easier to perform updates to dependencies by frequent smaller updates rather than large version jumps. By sticking close to the current upstream releases, any breaking changes are easier to deal with when they are smaller in number.

In the specific case of Jest, the following points are also relevant:
1. It was possible to simplify the configuration a bit (at least IMO).
2. I have removed the snapshot update in the `test` script. The snapshots are a kind of regression test. If we update them on each test run, they won't serve much function.
3. Removed the use of cross-env. As we don't need to specify `NODE_ENV`. The defaults should work fine.
4. Removed the need for our own `jest_config.js` for loading CSS files in tests. Although the file was trivial, it is one less file we have to maintain.

I hope this clarifies things a bit. Let me know if there is anything else.

I guess I should also go about having this PR rebased.

ranjib:
@Wraul lets keep dependency upgrades to dependabot, and do dedicated PR reviews only for issues. If there any feature/bug thats resolved by dependency upgrade, im happy to review those as PR. Otherwise lets just wait for dependabot's schedule

PR #1574, opened 2021-09-16
Make macro and steps logging consistent

revoohc:
Make macro and steps both log with macro-subsystem.  Makes
logging consistent and easier to parse.


ranjib:
thank you

Issue #1518, opened 2021-08-07
Input validation messages - translation


schwabix-1311:
To simplify the translation of the messages shown by input validation, I'd suggest to make them more generic.
E.g. if "This entry must be numeric" is used instead of "The upper temperature limit must be a number" this message would be usable for all numeric fields.
Since the message is shown next to the input field, the relation would be clear.

This would reduce translation effort tremendously.

schwabix-1311:
I started to work on this. Is there a way to self-assign?

And a generic question: some languages' grammar does not work with the string concatenation used e.g. in confirmation dialogs.
In English the message "<p>This action will delete {doser.name}.</p>" is fine. To allow translation it would ave to changed to something like <p>{i18n.t('doser:confirm_delete')} {doser.name}.</p>"
But German (possibly others languages too) grammar requires the verb at the end of this: "Dies wird OsmoDosierer löschen."

Would it be acceptable for translators to accept variable names enclosed in curlies inside the string? 
For instance  "doser:confirm_delete,Dies wird {doser.name} löschen."
Alternatively this could be solved by a String.format() as described here https://stackoverflow.com/questions/610406/javascript-equivalent-to-printf-string-format/4673436#4673436. This is more robust when var names change, but still the .csv file would include some strange syntax (perceived by a non-programmer).
Any other idea?
BTW, this is not limited to confirmation dialogs. Validation messages for minimum or maximum values could use "... ${min} ...", which is supported by Yup.

jaz1968:
The same problem with Balto-Slavic languages like Polish which is example of fusional language.

ranjib:
@schwabix-1311 while i like the idea, it will be hard to have this as a generic stance. There can be cases where we want a more specific sentence. The case you mentioned is reasonable. I suggest you take one case at a time and raise PR accordingly.

schwabix-1311:
PR1575 contains several alternatives for the delete confirmation dialogs.
1. I didn't get the documented way of replacing variables by i18next to work -> [https://i18next.github.io/i18next/pages/doc_features.html], keyword "replacing variables"
2. My preferred solution does not get jest's approval. This would be the placeholder "${somename}" in strings combined with a `i18n.t("key").replace("${somename}",actualvar)`. Jest's concern is very likely the missing backticks for string template interpolation, which in this case is not the intent.
3. Accepted by jest and working is $[somevarname] in strings, and the corresponding .replace(...)

[...]

PR #1575, opened 2021-09-17
Validations

schwabix-1311:
Issue #1518.  NOT YET complete.
However, before I invest too much time, I'd want to get feedback whether this direction gets your support.

Each commit covers validation of one main page. This includes the changes to translation in en/de, other languages are beyond my skills.
Confirmations for delete operations, plus calibration dialogs are also included.

And one fix I could not resist:  on Drivers page, the 'rpi' was presented differently, this is addressed.

There's about the same amount, may be a bit more in the pipeline.


schwabix-1311:
Why does jest complain?
The offending lines are all the same pattern, mimicking string interpolation in combination with  strings from i18next:
 `confirm(i18n.t('ato:title_delete').replace('${name}', probe.name), { description: message })`
with:
`ato:title_delete,"Delete ${name} ?"`
Do you have a better solution?

schwabix-1311:
Well, the pseudo string interpolation using  '$[name]' does work and is accepted by jest, while the form '${name}' is rejected.
The later is closer to real interpolation, therefore I'd prefer this.
What's your preference?

ranjib:
@schwabix-1311 can u isolate the changes, else it will harder for me to review. I think , some of the changes contains unnecessary line swaps, not sure why. Please raise the translations, delete validation and typo fixes as separate PRs. 
thank you

schwabix-1311:
The line swaps allowed me to use a diff tool between two main pages with similar structure & features (I don't remember which two these were) and thus to understand why they reacted differently.  The swaps can be undone, but honestly,  if I try to get this cleanly into git I will screw things up, almost guaranteed.

PR #1581, opened 2021-09-29
Issue 1557

schwabix-1311:
This fixes issue 1557. 
One change is unclear: front-end/src/timers/target.jsx, line 173 - is `state.lights` correct or `state.lightings` ? I found no reference to either.


ranjib:
in redux store its stored as lights : https://github.com/reef-pi/reef-pi/blob/master/front-end/src/redux/store.js#L14
subsystem is called lightings (in go side).
but this is complicated issue, the lighting backend subsystem still can not do this, all lights run in one thread and as a result macro enable/disable will not have any effect, till lighting subsystem is reloaded. we have to first address that issue (run each light in their own thread, and enable -disable will stop/restart light specific thread), before fixing the front end side of things

schwabix-1311:
Still not familiarized with redux and most of the go code I'll leave it up to you.
Please change front-end/src/timers/target.jsx#L173 as required, and when its time. The UI side should be OK now.

I did test the changes and in my dev build the timer successfully disabled and enabled the single light I have setup. Is your concern possibly only an issue with multiple lights?

PR #1607, opened 2021-11-01
UI care

schwabix-1311:
Various front-end improvements, each in a separate commit with explanatory comment.
Pick what you like (hopefully all)


schwabix-1311:
github through browser and my local git command line seem to be out-of-sync. I have no clue how to fix this.  
As a result in browser some very old commits keep re-appearing.

ranjib:
@schwabix-1311 start fresh (new branch) from master and cherry pick just the commits you want from the list, and then create a new PE from that branch, close this one,

I you want to clean up this one, then do the same as above followed by deleting the local ui_care branch, then rename the clean branch as ui_care , and then force push ui_care

schwabix-1311:
Besides the sync issue with github, jest fails in temperature_schema.test.js, although the UI seems to behave exactly as expected. The updated temperature_schema.js is not yet comitted; the fails are reported by a local 'make jest'.

 FAIL  front-end/src/temperature/temperature_schema.test.js
  ● Validation › should be invalid when heater and cooler are the same

    expect(received).toBe(expected) // Object.is equality

    Expected: false
    Received: true

      58 |     expect.assertions(1)
      59 |     return TemperatureSchema.isValid(tc).then(
    > 60 |       valid => expect(valid).toBe(false)
         |                              ^
      61 |     )
      62 |   })
      63 |

      at front-end/src/temperature/temperature_schema.test.js:60:30
Updated test code is attached
[temperature_schema.test.js.txt](https://github.com/reef-pi/reef-pi/files/7470513/temperature_schema.test.js.txt)
.

schwabix-1311:
Tried the 2nd approach to cleanup ui_care. I think I succeeded.
However, one jest issue in details line 557 persist. I have no idea what's wrong here. Can you have a look?

ranjib:
@schwabix-1311 update the test code (temperature_schema.test.js) to print the yup error message. that should show the specific fields which is causing the validation failure. 
https://www.techzaion.com/validation-with-yup

ranjib:
🙇‍♂️

PR #1636, opened 2021-11-03
(feat)add ph usage chart similar to temperature module

ranjib:


schwabix-1311 (commented):
LGTM
The backend already collected this data?
To have orthogonal features, a Reset button and corresponding Rest API would make it perfect ;-)

Do you expect me to approve?

ranjib:
yeah. I had refactored the backend to have same homeostasis library (telemetry included) across ph and temperature. UI work was left. There are whole bunch of UI enhancements that needs to be done to expose all the existing backend feature.
reset is a good call out, I'll get going on that.

PR #1668, opened 2021-11-21
fix: ph controller does not allow 'nothing'

schwabix-1311:
When one of upperFunction or lowerFunction is 'nothing' the controller bails out.
No using '' instead of 'nothing'. tcs seems to work correctly with heater/cooler='nothing'.


ranjib:
🦃

PR #1669, opened 2021-11-22
Ph usage chart translation updates

schwabix-1311:


schwabix-1311:
For symmetry I'd suggest to either 
add same separation of (history) and (history&usage) to temperature charts,
OR
give up pH(history) in favor of pH(history&usage)

What do you think?

ranjib:
thank you

PR #1689, opened 2021-12-06
(bug)dont save nothing as equipment id

ranjib:


schwabix-1311:
You can blame me for this breakage - sorry.
Why didn't thos show on my setup? I tested both, the dev build, and once 5.0 was released updated my Raspi. I have no cooler configured, does this make the difference?

PR #1698, opened 2021-12-28
ATO validation

schwabix-1311:
Change all validation messages to a common set reducing the translation effort.
Rearrange UI elements to be more intuitive:
- one-shot is affected by control, looked like it depends on warning
- maxAlert is the most important setting for warning, thus next to notify on/off
- changed default times for period and maxAlert
- moved button Save above chart, similar to other pages


front-end/src/ato/ato_schema.jsx
```diff
@@ -1,31 +1,38 @@
...
-    .integer()
-    .typeError(i18next.t('ato:chk_freq_number'))
-    .min(1, i18next.t('ato:chk_freq_number_value')),
-  pump: Yup.number(),
+    .required(i18n.t('validation:number_required'))
+    .integer(i18n.t('validation:number_required'))
+    .min(1, i18n.t('validation:integer_min_required')),
+
+  control: Yup.string(),
+  pump: Yup.string()
```
ranjib, line 17: @schwabix-1311 why pump is string? this is an ID

ranjib:
@schwabix-1311 can you please isolate the translation changes from aesthetics or other UI issues. e.g. the pump id changes,   Its easier for me to review this way.

schwabix-1311:
> @schwabix-1311 can you please isolate the translation changes from aesthetics or other UI issues. e.g. the pump id changes, Its easier for me to review this way.

The pump change is not aesthetics, it was meant to use similar patterns for similar fields. The four bullet points above ARE aesthetics / UI improvements, that I thought are small enough to include here while I touch this.

As a side note: meanwhile I spent so much more time on what I thought would be a simple translation (as reef-pi is advertised as being available in 10 or so languages), I'm getting tired. If you don't want this to be completed, please be candid and say it frankly. I could handle clear text better than wasting days and weeks and receive one reject after another. 
I am trying to fix a promise the code does not stand up to. You will get the credit for it. This is absolutely ok, that's the world of contributing to other's open source projects, but I would expect at least a bit of support by answering email, and/or fixing minor issues in PRs.

ranjib:
@schwabix-1311 im sorry you feel that way. These are not rejections, these are PR feedback, and please take it with heart. Your credits are amortized as the commit author, and I generally call out the contributors in r2r threads on point release.

 Your perception of opensource project is not correct.  There is no promise of 11 languages, there is only support for it, and that too from community contributions. I expect all of that to change and evolve over time. As your work and comment demonstrates, its harder to make it better (and it gets harder as the project grows) while its very easy to rant. 

I am sad that at the end you did not address the original comment, which would have made reviewing easier, patching easier and you a better programmer/collaborator.

Im merging this, as its new year, as I understand from where you are coming from, as I know you have good intention. I hope you learn. But know that, if you are looking for easy gratification, it will be hard for reef-pi which deals with significant complexity, risk, and smaller dev community.

PR #1699, opened 2021-12-28
Doser validation

schwabix-1311:
Change validation messages to a common set reducing the translation effort.

Completed translation of modal calibration.

Replaced calls to obsoleted showAlert/clearAlert.


ranjib:
this pr has common commits from the ato validation pr

schwabix-1311:
The ato validation PR introduces all these common validation messages. It is therefore the base for all validation PRs.
git-wise all *_validation branches are based on ato_validation.

If this is an issue I would have to split this PR once more, one for just the messages, and another for the ato specific changes. Then all *_validation branches including ato would have the same dependency.

ranjib:
What you have to do is to raise the first PR , get it merged and then raise the others, after rebasing from the new master. In its current form I have to scour through 17+ files for every PR based on that commit and keep a mental log of whats coming from previous PR, which is not optimal.

schwabix-1311:
The two commits all other PRs are based on are the top two, both clearly described as ATO. A visual tool like gitg would show this clearly.  And agreed, there's always room to improve. Next time.

ranjib:
rebase

ranjib:
thank  you :0)

PR #1701, opened 2021-12-28
Equipment validation

schwabix-1311:
Changed all validation messages to a common set reducing the translation effort.

Completed UI translation.


front-end/src/ato/ato_form.jsx
```diff
@@ -21,10 +21,10 @@ const AtoForm = withFormik({
       inlet: data.inlet || '',
       one_shot: data.one_shot || false,
       disable_on_alert: data.disable_on_alert || false,
-      period: data.period || 120,
+      period: data.period || 60,
```
ranjib, line 24: why?

ranjib:
@schwabix-1311 not sure what is the rationale of changing the default poll period. separate that out. please do a functional tests (it will be awesome if we can add any assertion on the smoke testing side). Rest looks good to me

schwabix-1311:
The idea of changing doser.period and doser.maxAlert was to show the user a reasonable relation by the suggested preset.
maxAlert=0 is definitely not. 
maxAlert<=period is bad too, since  every activation would trigger an alert. This relation would change with the distinction idlePeriod vs. activePeriod, which I still think would make a lot of sense and simplify the setup of an ato controller. Once I'm through the translation stuff I will try this and propose it as a PR.

I do not understand what you mean by "do a functional test ... add any assertion ..".  This PR did run through jest and smoke (smoke is difficult on a non-US locale! Plus my Linux distro's chrome browser is 'chromium', introducing yet another challenge to run smoke. A testcafe run against 'all' instead of 'chrome' would solve this part).

What kind of assertion would you like to see?  A sane magnitude of maxAlert WRT period could be asserted through yup though.

front-end/src/ato/ato_form.jsx
schwabix-1311: sorry, this question is ambiguous.

_Why is ato stuff changed **here**?_    a result of the branching I chose, see in equipment_validation.
_Or why is ato timing changed **at all**?_    see explanation below.

ranjib:
I understand the difficulties with integration /functional test. Dont worry about that right now, I'll see if I can add some coverage. By functional test I mean a coded assertions for those two values you mentioned (i.e ATO UI should raise warning if alert values are lowers than interval period ...). It could be something different, but as long as we get some confidence that it will not break the UI . anyhow.. don't worry about it now.. this looks good, except the internal time parts, I have to cross check that

front-end/src/ato/ato_form.jsx
ranjib: the later :-) , why change it to 120

PR #1703, opened 2021-12-29
Lighting validation

schwabix-1311:
Changed all validation messages to a common set reducing the translation effort.
    
Completed UI translation.


front-end/src/lighting/main.jsx
```diff
@@ -196,6 +187,21 @@ class main extends React.Component {
     }
   }
 
+  getModeLabel (mode) {
+    let label = ''
```
ranjib, line 191: what if you remove the variable initialization and break statement all together, and just return on every case
```js
switch(mode) {
case 'auto':
  return i18n.t(...)
case '':
 return ...
} 
```

PR #1731, opened 2022-01-05
Misc validation

schwabix-1311:
Misc validation translations:  configuration settings, camera
Added few UI translations for health chart type, display settings


ranjib:
🏆

PR #1735, opened 2022-01-07
 Instance validation & missing translations

schwabix-1311:
Generic validation messages & a few translations.

**Where is <Instance ...> used?**  
My hack to(temporarily) enable it, revealed
errors in the bound function names - fixed that, w/o a chance to test,
since the hack enabled the tab, which then showed several 404 errors.

### DIFF
--- a/front-end/src/drivers/main.jsx
+++ b/front-end/src/drivers/main.jsx
@@ -38,7 +38,7 @@ class drivers extends React.Component {
             driver={d}
             validate={this.validate}
             driverOptions={this.props.driverOptions}
-            read_only={(d.type === 'rpi')} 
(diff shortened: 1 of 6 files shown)


ranjib:
instance was intended and manager were two component that were intended to be used for cluster support (multiple reef-pi control from a single place). Its not being developed now, and nowhere used. I have to delete it.

schwabix-1311:
Should I close this? Or do you want to keep as reminder for deletion of instance module?

ranjib:
@schwabix-1311 i'll merge this. thank you. I want to get back to this (supporting multiple reef-pi from a single place)

PR #1737, opened 2022-01-09
Macro validation + translations

schwabix-1311:
Changed to use generic validation messages.

Added translation of subsystems, which hopefully doesn't break functionality.

The new string group 'function:*' will be used again (and extended) by the
coming PR for timers.
Long term I'd want to consolidate the subsystem names as used on tab titles,
checkboxes of Settings-Capabilities, and in macros and timers to the two
string groups:
- capabilities:* for the category name (tab and caps.) and
- function:* for the individual controlled item.
This will result in a more consistent terminology throughout the UI.


front-end/src/macro/generic_step.jsx
```diff
@@ -76,13 +76,15 @@ const mapStateToProps = state => {
...
     subsystem: [
-      { id: 'timers', name: 'timer' },
-      { id: 'phprobes', name: 'pH' },
-      { id: 'ato', name: 'ATO' },
-      { id: 'temperature', name: 'temperature' },
-      { id: 'lightings', name: 'lighting' },
-      { id: 'system', name: 'system' },
-      { id: 'doser', name: 'doser' }
+      // capabilities:..  are the subsytem names (plural or cathegory), corresponding to tab pages, whereas
+      // function:... are individual devices in these cathegories
```
ranjib, line 80: individual entities? Things like macro, timer are not devices.

schwabix-1311: You're right, 'device' may not be the best term (non-native speaker ...).
The idea is to clearly distinguish the containers (visible as tabs and settings:capabilities) from the controller entities.

This may be the preparation to decouple containers from entities, such that the user can collect all his analog probes on one tab, and all things reading digital inputs on another, by setting the container name himself.
The difficulty here is the REST API, as this should use fixed paths I guess.

PR #1739, opened 2022-01-14
Timers validation and UI

schwabix-1311:
Changes to use generic validation messages.

Added translation of system names, similar to previous PR for macros, string group "function:*".

Swept obsoleted timer strings.

Fixed an issue with initial UI state of fields 'on'/'revert'/'duration'. For equipment this seems fixed by a changed variable initialization.

- [ ] FIXME: other similar functions/entities still broken, see FIXME in edit_timer.jsx! Either these functions shouldn't show the affected fields, or their variable initialization is missing.

- [ ] Unsolved issue: timers:running vs. :run, and :reverting vs. :revert never seen on UI, likewise 'in-use' for connectors. With some debugging I found for instance props.equipment to be undefined at the time 'in-use' would be made visible. Did not spend time on the 'reverting', but assume a similar root cause as for 'in-use'.

AFAIK this is the last PR for validation translations. **Hurray!**


ranjib:
in-use is another unfinished work. it was supposed to enable us checking associated entities with another entity. i.e. if driver is being deleted, check all connectors, equipment associated with it is deleted first. i didn't get to finish it yet.

thank you for the hard work. validation messages looks much better now

Issue #1842, opened 2022-05-21
Page title under a small window doesn't match actual title


richievos:
The title being rendered on mobile / small screens is the name of the `cap` key versus the actually translated text.

1. open up the app on a small screen
2. navigate to something like Log
3. close the hamburger menu

Result: the text in the header is "log"
Actual: the text in the header is "Log"

This is pretty minor, but just noting it because I tried to fix it and I'm too bad at React to figure it out. What I really wanted to do was set the `document.title` to `reef-pi | {page title}` and set the header title to match, but I'm not sure how to do that in React.

It might be more confusing if I were using a different language and the translations wasn't the same as the name of the property in `caps`

Issue #1780, opened 2022-03-10
[UX] Use Deep Links in Header


richievos:
I commonly want to pop open multiple tabs of reef-pi things. For instance, seeing the dashboard while messing with ATO settings, or pop open ATO+Dashboard+Configuration. Right now because they aren't actually links, I have to dup the window and then navigate to them. I'd normally expect to just be able to right-click and "Open Link in New Tab" to pop open another one.

Generally being able to deep-link in to reef pi would be helpful.

Given this is a React app I think this is just a matter of some push state hooks or integrate switching to React Router.

Thoughts?

ranjib:
i am ok to support this as long as it does not break the current single page user experience.

ranjib:
thank you @richievos for addressing this. its in main branch now

PR #1843, opened 2022-05-22
Support direct links to pages

richievos:
This replaces the redux / custom management of the pages with a React Router
based version. The main impact of this is now you can directly go to urls
ala `/temperature` and it will render the appropriate page. The nav links
also point to this, which means you can also "open in new tab" to have
multiple reef pi tabs working at once.

I have tested this by going through the various pages and adding/deleting
things, playing with the forms, and navigating to various pages. It all
works for me. However I'm not sure if there's existing precedent for auto
testing this sort of change in this app.

I would recommend poking around at it a bit yourself as well to triple
check I didn't miss some behavior change, but it worked in my testing.

Implements #1780

Note while implementing this I found #1842, which I didn't try and fix in
this commit. Mostly because I'm not sure how to implement it in React, and I
gave up trying for the moment.

### DIFF
--- a/front-end/src/configuration/configuration.test.js
+++ b/front-end/src/configuration/configuration.test.js
@@ -26,7 +26,6 @@ describe('Configuration ui', () => {
 
   it('<Main />', () => {
     const m = shallow(<Main store={mockStore()} />).instance()
-    m.setBody(1)()
   })
 
   it('<Admin />', () => {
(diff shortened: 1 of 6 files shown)


richievos:
I added a second commit to this PR doing the same deep linking for configuration sub-tabs. That isn't needed to merge the first commit, but it's part of the same feature so I'm putting them on a single PR.

ranjib:
:bow:

Issue #1779, opened 2022-03-10
[UX] Errors are hard to find


richievos:
When Errors occur, actually viewing them is difficult unless you already know they're under Configuration->Errors. As a reef-pi newb, I was having issues, and saw Errors(3) in the footer. I clicked Log, assuming they'd be in there, and saw it was empty. I even looked in my developer web console thinking maybe they were in there somehow, and couldn't find them.

I only stumbled upon them under Configuration because I was looking at something else (the other issues I filed).

Suggestions:
A. [merge] Merge Errors into Log. As a user I don't think there's an intuitive reason they'd be separate, even if in implementation there's nuances.
A.1. Embed them inline with others
A.2. Show them on that page, but separated
B. [link] Make the Errors(##) text a link, which takes you to them. I assume the main complication with that is the difficulties of navigating, but possibly A helps with that if it's the same as clicking Log

I think the _link_ suggestion would be likely more immediately beneficial than the merge one, even if the merge one was argued to be a nice setup.

ranjib:
log was built to  capture only UI specific, client side issue and not backend 5xx/4xx errors. I agree its confusing. I think its a good idea to move the current configuration -> error section to Log, and rename it as just "error", and link it from footer section. This will involve some changes, and i think if its API breaking we can ship it as part of 6.0. if this does not require api breakage, we can ship it as part of any 5.x release

ranjib:
https://github.com/reef-pi/reef-pi/pull/1893

PR #1844, opened 2022-05-22
Navigable errors link

richievos:
Link the Errors in footer to the Errors page

This is built on top of #1843. This doesn't combine the Log and Error screens,
but instead deep links to the error page.

I believe combining the Log and Errors screens will still be worthwhile, but
this is an easy improvement with the deep linking in place.

Tested just by pulling up the UI and seeing the link works.

Relates #1779

**Merge #1843 first**


richievos:
I had a failing test due to deleting a no longer needed method. I'm not sure what was the intention to that test so I did a very minor change to unbreak it.

c6d534a5ced3f40a885b0d9bf9851ad5eb5e0ce0

Issue #1846, opened 2022-05-23
Configuration screen sometimes shows erroneous setting values


richievos:
While doing more QA of #1843 I found that the configuration on first render (if you go to http://reefpi/configuration) is empty. I have been walking through the state tracking to figure out how to fix that, and the gist is the nested forms and checkboxes aren't getting refreshed when the data finishes fetching. This wasn't a problem previously because by the time you clicked through to that page, it's likely the config would've already loaded. However it's now obvious.

I then went back to my existing pi to just understand how it handles it, and I realized there's actually similar bugs in the released version. What I've found is if I click around the different configuration screens, and keep going back to the main settings page, sometimes the checkboxes aren't set properly. In fact, on first render almost always (or maybe always) the top checkboxes are all unchecked.

**Repro on release:**
1. go to reef pi
2. click Configuration
3. *Note how Notification, Enable profiling, ... are all unchecked*
4. Click to Errors (or anything else)
5. Click back to Settings
6. *Note how whatever ones you expected to be checked, such as Notification, are now checked*

**Repro on master:**
1. go to http://reefpi/configuration
2. *Note how all the boxes are unchecked*
3. Click to Errors (or anything else)
4. Click back to Settings
5. *Note how whatever ones you expected to be checked are now checked*

I'm working on a fix, but if you have an immediate knowledge of how to get those to re-render after Redux has properly loaded, LMK. I'm a bit out of my React depth on this. My normal approach would be to have it just try and not render at all until the loaded call finishes, but I am not sure an idiomatic way to do that.

richievos:
My workaround with the loading text I think fixed this. That seems _good enough_, so closing this

PR #1847, opened 2022-05-23
Add more data loading to config screens

richievos:
Two separate commits fixing separate aspects of the connection screen data loading. These are technically issues even if you went to `http://reefpi/` and then quickly navigated to the other pages, but being able to jump to `http://reefpi/configuration/*` exposed it.

Relate to #1846


ranjib:
:hugs:

PR #1848, opened 2022-05-24
Don't require a heater or cooler temp when they're disabled

### LINKED ISSUE #1836: Temp hysteresis validation errors when cooler disabled
This is similar to #1786 and #910, but requires a couple steps to get yourself into a bad state. The problem is you can get the heater temp = the cooler temp by enabling/disabling them incrementally, and then you're stuck and can't edit and it's not clear why.

Series of steps:

## Setup
1. create a temp setup
2. enable the heater control, set it to 79 (pick whatever equipment just to get it saved, or a fake macro)
3. enable the cooler control, set it to 80 (pick whatever equipment just to get it saved, or a fake macro)

## Trigger the first issue
4. flip the cooler control to None
5. change the heater temp to 80

You'll now get a "Hysteresis is less than acceptable threshold" error and can't save.

## Trigger the second issue
6. flip the heater control to None

You now can save. This will result in heater value = cooler value = 80)

## Trigger the final issue
7. Re-enable the heater

You won't see any validation errors

8. Hit save

You'll get the hysteresis error again.

That's somewhat expected if you know what's going on, but you'll not have any reason why you're actually getting that error.

# Proposed fix
I'm still not exactly sure what hysteresis does honestly, but generally if only one of cooler/heater is enabled then the temp values for the other one should be completely ignored.

richievos:
Fixes #1836

TBH I'm not sure exactly what the behavior of hysteresis is, so I'm not
sure if there's cases where this validation is erroneous. However, this
fixes the bug where right now if you try and create a new temperature
with just a heater, it will fail validation without telling you why. Currently
it's requiring you to setup a cooler regardless, as outlined in #1836

### DIFF
--- a/front-end/src/temperature/temperature_schema.jsx
+++ b/front-end/src/temperature/temperature_schema.jsx
@@ -103,6 +103,8 @@ const TemperatureSchema = Yup.object().shape({
           .typeError(i18n.t('validation:number_required'))
           .test('lessThan', i18n.t('temperature:hysteresis_less_than'), function (hyst) {
             if (this.parent.heater === undefined || this.parent.heater === '') { return true }
+            if (this.parent.cooler === undefined || this.parent.cooler === '') { return true }
+            if (this.parent.min === undefined || this.parent.max === undefined) { return true }
             return hyst < (this.parent.max - this.parent.min)
           })
       } else { return schema }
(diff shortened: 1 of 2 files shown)


ranjib:
:hugs: :angel:

Issue #1830, opened 2022-05-02
fix spurious timer validation error


ranjib:
```
Something changed with build 5.1 or 5.2 where setting up the dosing timers will not longer accept */4 to run every 4 hours. I get the error.



[ An integer, a range (e.g. 1-5) or * is required! ]



I am having issues figuring out the new requirements for scheduling. It used to be in cronjob format, now it's not?

```

https://www.reef2reef.com/threads/reef-pi-an-opensource-reef-tank-controller-based-on-raspberry-pi.289256/page-921

ranjib:
we have removed Ui based validations in lieu of backend based validation. this is to simplify the cron expression validation logic

PR #1850, opened 2022-05-25
Add a failing test for a basic cron of every second

richievos:
Relates #1830

----

This doesn't fix the linked issue, but repros it. Hopefully this is helpful for a subsequent fix.

### DIFF
--- a/front-end/src/doser/doser_schema.test.js
+++ b/front-end/src/doser/doser_schema.test.js
@@ -1,0 +1,55 @@
+import DoserSchema from './doser_schema'
+
+describe('DoserValidation', () => {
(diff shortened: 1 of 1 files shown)


ranjib:
@richievos it was this https://github.com/reef-pi/reef-pi/commit/4f1ece4d54972125240af1cd0fcf6acba9fe2dde 
can you just remove the regex based validation in lieu of just string, like previous.
I have cross checked, we have backend validation for create, update API calls so its not a big risk. 
much appreciated

ranjib:
https://github.com/reef-pi/reef-pi/pull/1851

PR #1852, opened 2022-05-26
Fix doser regex

richievos:
This PR has a couple options in it. The final version is just deleting the regex completely, which I'm glad to see @ranjib suggested in #1850 because that seems like the most accurate solution.

87e3563e8c6ac0e0dc3e64f1d436389ad97ac30d updated the regex to support the more complicated (but not all) cron variants.

I also cleaned up a couple test warnings.


ranjib:
:bow:

Issue #1882, opened 2022-07-04
Cannot disable Adafruit.IO without entering in Email Alert password


richievos:
There's a validation bug where trying to disable Adafruit.IO leads to the Email alerts trying to get updated, requiring you to re-enter your password. Because of this there's no way to disable Adafruit without reentering your other passwords/info.

1. go to /configuration/telemetry
2. setup Adafruit.IO and Email alerts
3. hit Update
4. Reload

Now your reef-pi has multiple telemetry syncs

5. go back to /configuration/telemetry
6. Uncheck Adafrui.IO
7. hit Update

**Expected behavior**: Telemetry settings update, leaving my email alerts settings alone and disabling Adafruit.IO

**Actual behavior**: You get an error saying "Please set a valid mail password" and the update doesn't go through. You therefore can't disable Adafruit.IO without re-entering your email password. That's somewhat annoying generally, but more of a PITA if you're using 2fac auth with gmail and app passwords. That means you need to regenerate a new app password.

I assume this same issue exists for other types of Telemetry changes, such as if you have Adafruit.IO enabled and you try and disable Email alerts.

**Workaround**: enter in a password under Email alerts before hitting Update

**Fix**: obviously don't do that. Ideally if I don't change any credential text it shouldn't think I need to re-enter them. I'm not really sure the normal UI treatment this type of thing would get in other apps to do that. As a spot-fix for this, allowing an empty password under Email alerts would fix it, but that'd lead to people forgetting to fill it in when editing. However that seems like less of an issue then disable being completely broken.

richievos:
RFC diff at #1883

PR #1883, opened 2022-07-04
RFC: Allow saving telemetry without reentering creds

richievos:
After you have persisted telemetry settings, the UI no longer knows your creds.
This is good in that it avoids saved passwords getting returned out of the
backend, but it causes the UI to only allow updating telemetry if every time
you reenter in all the relevant passwords/tokens. That makes it very hard to
edit these later.

This is caused by the API blanking out those fields before returning them. That
is reasonable. However that leads to two issues. The first is the UI's
validation that token/password are filled in fails. The second is that if
you made the call to the backend, it'd then cause the database to be written
without those creds in it, losing them.

The current fix could be improved, as noted in the TODOs. However, given
right now the update flow has a major usability issue, I feel this is a
reasonable short-term solution. A future one likely would involve having the
API understand how to do an Upsert and ignore any fields that are not passed
through on the JSON. Then the UI would only pass these in if they changed,
or there'd be some sort of separate credentials form submitted separately.

Relates #1882

# This PR not yet tested, I'm sending for feedback/opinion

Before vetting this further, I want to get an opinion on this solution from @ranjib. I'll give it some testing locally on my side because this'll solve an issue I'm currently getting, but before cleaning anything up or going further I wanted to get a general  👍.

### DIFF
--- a/controller/storage/store.go
+++ b/controller/storage/store.go
@@ -49,3 +49,9 @@ type Store interface {
 	CreateBucket(string) error
 	Path() string
 }
+
+type DoesNotExistError struct { text string }
+
+func (e DoesNotExistError) Error() string {
+	return e.text
+}
--- a/controller/telemetry/api.go
+++ b/controller/telemetry/api.go
@@ -5,7 +5,15 @@ import (
 	"time"
 
 	"github.com/reef-pi/reef-pi/controller/utils"
+	"github.com/reef-pi/reef-pi/controller/storage"
 )
+
+// TODO: translate these
+// TODO: this is a bit of a hack in that it means someone can't have their
+// 			 actual token/password be the string "<stored>", but that seems rare
+//			 enough a reasonable trade-off for a quick fix to unsaveable form bugs
+const PasswordStoredPlaceholder = "<stored>"
+const AdafruitIOTokenStoredPlaceholder = "<stored>"
 
 func (t *telemetry) GetConfig(w http.ResponseWriter, req *http.Request) {
 	fn := func(_ string) (interface{}, error) {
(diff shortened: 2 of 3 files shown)


richievos:
I did some tweaks to get this to fully compile (caveat I have very little go experience). I also did some initial testing. This solution seems to work, though I need to fully QA it end-to-end, and ideally add some tests, before saying it really works.

controller/storage/store.go
```diff
@@ -49,3 +49,9 @@ type Store interface {
 	CreateBucket(string) error
 	Path() string
 }
+
+type DoesNotExistError struct { text string }
```
ranjib, line 53: model it like this
```go
var ErrDoesNotExist = errors.New("entitiy does not exist")
// for returning this type of error (name is a string variable)
return fmt.Errorf("entity name:%s error:%w", name, ErrDoesNotExist)  // creates an error of ErrDoesNotExist family, by wrapping it
// for checking if an error belong to this family (err is returns through fmt.Errorf()
if errors.Is(err, ErrDoesNotExist) {
 . ///
}
```

controller/telemetry/api.go
```diff
@@ -5,24 +5,57 @@ import (
...
+
 func (t *telemetry) GetConfig(w http.ResponseWriter, req *http.Request) {
 	fn := func(_ string) (interface{}, error) {
 		var c TelemetryConfig
 		if err := t.store.Get(t.bucket, DBKey, &c); err != nil {
 			return nil, err
 		}
-		c.AdafruitIO.Token = ""
-		c.Mailer.Password = ""
+		if (c.AdafruitIO.Token != "") {
```
ranjib, line 24: @richievos  i think we can get rid of the behavior of returning telemetry config with the password scrubbed completely.. and remove this special placeholder logic. After 6 years of use i find it more pain than a security feature

ranjib (commented):
@richievos left some minor nits (around error modeling) . I think its best to just remove the resetting of token/passwords in the GetConfig method and not introduce any place holder logic. What you think?

ranjib:
@richievos closing this due to inactivity. Feel free to reopen if you want to continue working on this

richievos:
@ranjib  can you reopen this? I'm unable to. I didn't respond to your comment, but AFAIK this is still broken on main. I think at least having this open to record that there's a known bug would be helpful (I'm currently running this variant on my reef-pis).

controller/storage/store.go
richievos: done

richievos:
Stylistic tweaks addressed. Rebased on latest main

PR #1887, opened 2022-07-10
(chore)add warning message in footer when reef-pi is running on  dev mode

ranjib:


front-end/assets/translations/en.csv
```diff
@@ -14,6 +14,7 @@ clear,Clear
 columns,Columns
 configure,Configure
 delete,Delete
+devmode_warning,In DevMode
```
richievos, line 17: I'd recommend making this warning more detailed. Dev mode isn't very descriptive if you don't know the ins and outs of reef pi.

How about something like: "Warning Input/Output Disabled (dev_mode)"

ranjib: i thought about this.. but then it will become too long.. i worry it will overflow the area . i was thinking of an icon maybe..

ranjib:
@richievos merging this for now, will add the icon or a better text soon we find something suitable

PR #1890, opened 2022-07-11
(Chore) fix react router warning and duplicate route declaration

ranjib:
/cc @richievos


richievos:
Can you double check with this that all the links work? Particularly I feel I tried something like this and I left it the way it was because the deep links weren't working with it this way.

Generally I was verifying by just clicking the config links in the UI, and refreshing my page, and making sure the same thing came back. Unfortunately I couldn't think of a good way to test all that programmatically, at least without bootstrapping a big selenium style test...

I'd double check the main tab, and at least one of the nested ones.

Otherwise, lgtm 👍

PR #2015, opened 2023-01-30
Continue on partial failure in temp control

### LINKED ISSUE #2013: Temperature logic fails to control heater if cooler fails
I started getting really erratic issues with my heat control failing. The heater is never kicking on properly based on temperature, but I can always manually trigger it through the equipment setting. Long-story short, I figured out that the problem is I'd unplugged the kasa outlet that controls my cooler (a fan) and it's triggering bad behavior in the temp code.

I finally discovered what was going on by looking at the Errors, and seeing "Failed to execute temperature control logic for:Sump Temp F. Error:dial tcp 192.168.0.22:9999: i/o timeout". That was confusing, because the equipment was still working, and when I looked at my heater's smart plug ip it was not that. Eventually after more digging I found that `192.168.0.22` is the fan I'm using as the cooler on temperature. 

As a workaround I've disabled the cooler on my temperature control, and now it works again.

**Issues (not all verified)**:
* [verified] if your cooler fails, the heater will not turn on, even if you're heating
* [not verified] I assume this isn't unique to the kasa driver, but any driver
* [not verified] I assume the same can happen with a cooler, where a heater failing to control could prevent a cooler from activating

# Repro of my case
1. create two kasa drivers, call one HeaterDriver and another CoolerDriver
2. create two equipments, Heater=HeaterDriver, Cooler=CoolerDriver
3. create one temperature, Control=Equipment, Heater=Heater, Cooler=Cooler
4. trigger temp changes, causing the heater to kick on and the cooler -- everything should work

Now the bug:

5. break the connection to CoolerDriver (unplug the cooler plug, or make the ip for CoolerDriver bad, or ...)
6. cause the heater to need to flip on (cool the temp probe, change the heater target temp, ...)

**Expected behavior**: the heater kicks on, because the cooler doesn't matter

**Actual behavior**: the heater *never* kicks on, because of the cooler failing

richievos:
**WIP NOTE**
This is a WIP. I haven't tested it locally yet with my actual reef-pi, nor do I know how to write an automated test for it. I don't see any controller tests when I did a quick check.

Sending as a WIP to give a heads-up on the fix.
**/WIP NOTE**

This updates the control code to flip as many heater/coolers as possible into the correct state when performing an action, regardless of what happens to some of them. For instance, if we're trying to heat the tank, we should try and turn on the heater regardless of what happens to the cooler. Similarly that same change applies to flipping on a cooler or turning everything off.

It's generally important this tries to flip everything to the proper state versus just reordering things. For instance, if we're trying to heat the tank we should flip the cooler off regardless of what happens to the heater and flip the heater on regardless of the cooler.

Fixes #2013

### DIFF
--- a/controller/homeostatsis.go
+++ b/controller/homeostatsis.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"fmt"
 	"log"
 	"math"
 	"time"
@@ -95,6 +96,15 @@ func NewHomeostasis(c Controller, config HomeoStasisConfig) *Homeostasis {
 	return &h
 }
 
+// a very basic equivalent of errors.Join, but works for go < 1.2
+func BasicErrJoin(prevErr error, newErr error) error {
+	if prevErr != nil {
+		return fmt.Errorf("%w; %v", newErr, prevErr.Error())
+	} else {
+		return newErr
+	}
+}
+
 func (h *Homeostasis) Sub() Subsystem {
 	if h.config.IsMacro {
 		return h.macros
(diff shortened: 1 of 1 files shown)


ranjib:
we can wrap errors with fmt now, we dont need external mulltierror. just use fmt.Errorf("upper error: %w", err).

richievos:
@ranjib the behavior of fmt.Errorf is slightly different than what I'm doing here, at least I believe so based on my limited understanding of Go. I'm using multierrors.Join to combine multiple errors, because there may be an error from each step. I believe Go 1.2's `errors.Join` is the equivalent.

I'm pushing an updated diff that leaves go at 1.16, and I'll just manually implement an equivalent of `.Join` which can be easily replaced later.

richievos:
Rebased, removed multierrors, instead using a helper function. If you're cool with upgrading Go to 1.2 I can kill that helper and upgrade. I'd prefer to do that in a different commit / PR though.

ranjib:
Go 1.2 upgrade will be awesome

ranjib:
thank you @richievos

PR #2057, opened 2023-04-13
Fix telemetry creds flow

richievos:
After you have persisted telemetry settings, the UI no longer knows your creds.
This is good in that it avoids saved passwords getting returned out of the
backend, but it causes the UI to only allow updating telemetry if every time
you reenter in all the relevant passwords/tokens. That makes it very hard to
edit these later.

This is caused by the API blanking out those fields before returning them. That
is reasonable. However that leads to two issues. The first is the UI's
validation that token/password are filled in fails. The second is that if
you made the call to the backend, it'd then cause the database to be written
without those creds in it, losing them.

The current fix could be improved, as noted in the TODOs. However, given
right now the update flow has a major usability issue, I feel this is a
reasonable short-term solution. A future one likely would involve having the
API understand how to do an Upsert and ignore any fields that are not passed
through on the JSON. Then the UI would only pass these in if they changed,
or there'd be some sort of separate credentials form submitted separately.

Relates #1882

---
This is a duplicate of #1883, but I can't open 1883.

### DIFF
--- a/controller/storage/boltdb.go
+++ b/controller/storage/boltdb.go
@@ -42,12 +42,12 @@ func (s *store) bucket(tx *bolt.Tx, name string) (*bolt.Bucket, error) {
 	} else {
 		p := tx.Bucket([]byte(s.parent))
 		if p == nil {
-			return nil, &DoesNotExistError{fmt.Sprintf("Parent bucket: '%s' does not exist.", s.parent)}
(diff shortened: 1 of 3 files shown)


richievos:
@ranjib  regarding  your comments on the other PR, I think returning it scrubbed is valuable. If only because that's how every other site does it, and it'd ... feel weird if reef-pi didn't.

ranjib:
Thank you @richievos

PR #2087, opened 2023-07-22
Update reef-pi.service

BDeakins:
Added WANT and AFTER network-online.target.  Prevents the service from attempting to bind before the network has started.  Absence was causing the service to fail to bind when a specific IP was specified in the config.


ranjib:
Thank you, this was much needed

PR #2156, opened 2023-11-12
Update parse config

hectorespert:
`ioutil.ReadFile` is deprecated in favour of `os.ReadFile`


ranjib:
Thank you @hectorespert

PR #2233, opened 2024-08-16
Store hashed password

hectorespert:
Persist hashed password in the database instead a plain password to improve security

I choose the 'bcrypt' hash algorithms because it's easy to implement: https://snyk.io/blog/secure-password-hashing-in-go/


ranjib:
@hectorespert sorry for the delay. Merging this. I'll do some sanity test before we include it in the release

PR #2278, opened 2024-12-27
Raspberry not in dev mode by default

hectorespert:
After https://github.com/reef-pi/reef-pi/pull/1976 Raspberry driver doesn't work because is loaded in devel model


hectorespert:
Closed in favor of https://github.com/reef-pi/reef-pi/pull/2277

PR #2279, opened 2024-12-29
(chore) Capture errors on stop controller

hectorespert:
Capture error on stop controller when SIGTERM is received


ranjib:
thank you

PR #2280, opened 2024-12-29
(test) Add a light controller test

hectorespert:
Add a light controller test


ranjib:
thank you @hectorespert

PR #2301, opened 2025-02-09
(ci) Release Workflow

hectorespert:
Add a release workflow to automate the generation of release artifacts

This workflow is tested in my fork
- https://github.com/hectorespert/planted-pi/actions/runs/13224885524
- https://github.com/hectorespert/planted-pi/releases/tag/2025.1.1

### DIFF
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -1,8 +1,21 @@
 name: deb
-on: [push, pull_request]
+on:
+  workflow_call:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
 jobs:
-  builds:
+  build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


ranjib:
@hectorespert do you have documentation I can refer to for the release/download actions. Do we have to setup GitHub token? or every thing is automagical

hectorespert:
@ranjib  It's magic, `secrets.GITHUB_TOKEN` is automatically added by GitHub and doesn't require any configuration. https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

The `release` workflow listens a release `created` event and then starts to prepare the release artifacts. https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

I tested it generating a release using the Release UI.

Screenshot from 2025-02-24 18-57-07

PR #2490, opened 2026-04-10
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `controller/telemetry/mqtt.go:L49`

The MQTT client unconditionally sets `tls.Config{InsecureSkipVerify: true}`. This disables server certificate verification and allows man-in-the-middle interception or modification of telemetry data and credentials sent to the broker.

## Solution

Remove `InsecureSkipVerify: true` and perform normal certificate validation. Allow users to configure a trusted CA/certificate if needed, and only permit insecure mode behind an explicit, clearly documented development-only flag.

## Changes

- `controller/telemetry/mqtt.go` (modified)
- `controller/utils/credentials_manager.go` (modified)

### DIFF
--- a/controller/telemetry/mqtt.go
+++ b/controller/telemetry/mqtt.go
@@ -47,7 +47,7 @@ func NewMQTTClient(conf MQTTConfig) (*MQTTClient, error) {
 			connOpts.SetPassword(conf.Password)
 		}
 	}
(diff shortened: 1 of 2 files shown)


ranjib:
@tomaioo we have to make the mqtt configurable, can not enforce tls by default as well, raise a issue, discuss there and then send PR

PR #2499, opened 2026-04-11
feat: add ATO debounce delay

### LINKED ISSUE #1351: Can we have a toggle in the ATO section that allows a 10 second delay on the input float sensor?
Tunze has this feature to stop the ATO from cycling off and on when water level drops for a brief second due to a power head in the sump or sudden water level change from movement in the sump. I do not think that everybody would benefit from this feature so that is why I asked for a toggle.

ranjib:
Fixes #1351

## Summary
- Added `Debounce time.Duration` field to the `ATO` struct (stored as seconds, defaults to 0 = disabled)
- Backend tracks per-ATO `lowSince` timestamp in the Controller; pump only activates after the sensor has been continuously low for at least `Debounce` seconds
- Added "Debounce" number input field to the ATO edit form
- Fully backwards compatible: `debounce: 0` is the default and behaves exactly as before

## Test plan
- [ ] Create/edit an ATO — a "Debounce" field (in seconds) should appear in the form
- [ ] Set debounce to 10 seconds; verify the pump does not activate immediately when sensor reads low — it waits 10s
- [ ] Set debounce to 0; verify original immediate behavior is preserved
- [ ] All ATO Go tests pass: `go test ./controller/modules/ato/...`
- [ ] All ATO JS tests pass

### DIFF
--- a/controller/modules/ato/ato.go
+++ b/controller/modules/ato/ato.go
@@ -21,6 +21,7 @@ type ATO struct {
 	Inlet          string        `json:"inlet"`
 	Pump           string        `json:"pump"`
 	Period         time.Duration `json:"period"`
+	Debounce       time.Duration `json:"debounce"`
 	Control        bool          `json:"control"`
 	Enable         bool          `json:"enable"`
 	Notify         Notify        `json:"notify"`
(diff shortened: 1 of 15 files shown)


ranjib:
@copilot resolve merge conflicts

PR #2502, opened 2026-04-11
feat: auto-provision connectors when a driver is added

### LINKED ISSUE #1899: When adding a plug driver, auto add all the related peripherals
As a usability issue, it's tedious to setup equipment. As an example, setting up a kasa hs300 smart strip requires oodles of UI work.

1. Find your smart strip on your router, pin its ip, grab its ip
2. Go into reef pi, admin, drivers
3. Enter in the hs300, with the address
4. Admin->Reload (this may be unneeded, I can never remember)
5. Go to the connector screen
6. 6 times repeat:
     * Hit + under outlets
     * Enter in pin=0, select the driver, enter in a name like "blah pin=0, plug=1
     * hit save
 7. Admin->Reload (this may be unneeded, I can never remember)
 8. Go to equipment
 9. 6 times repeat:
     * Hit +
     * select the outlet, name in the exact same thing
     * Hit save

It's very tedious, and easy to screw up.

This should all be automated, and happen automatically when you add the kasa. It should create every related thing the kasa would have. All the inputs, all the equipment.

Outside of this not having been implemented yet, is there any reason it doesn't work this way?

Additionally, deleting the driver should also delete all these other things too.

As part of my creation on some python scripts to make my life managing my reef pi better I've done this. I made it so the script upsets all the relevant outlets/equipment when I add the driver. I could backport the same capability into the UI, but wanted to verify first.

The least confusing for a new user setup I believe would just have it automatically do this. The closest to the current approach version would be make it so there's a checkbox that defaults to checked saying "Add associated outlets and equipment" and if unchecked it'd act like now. There also could be a "sync related outlets and equipment" button that syncs everything again. That'd be particularly useful when you update the name of the driver, it'd sync everything again.

ranjib:
## Summary

- When a driver is created, reef-pi now automatically provisions all connectors exposed by that driver based on its capabilities: **DigitalOutput → Outlets**, **DigitalInput → Inlets**, **PWM → Jacks**, **AnalogInput → Analog Inputs**
- Individual connector failures are logged but do not abort provisioning (e.g. adding the same driver twice won't fail — already-existing connectors are skipped with a log message)
- Adds a `POST /api/drivers/{id}/provision` endpoint for re-provisioning at any time (e.g. to sync after renaming)
- Adds a **Provision** button to each driver card in the Drivers UI that calls the provision endpoint and refreshes the page

Before this change, setting up a 6-outlet Kasa HS300 required 6 manual outlet-creation steps plus 6 more equipment steps. Now it's automatic on driver creation.

## Test plan

- [x] `go test ./controller/device_manager/...` — all tests pass
- [x] `go test ./controller/...` — all tests pass
- [x] Frontend tests: `npm test -- --testPathPattern="connectors|driver"` — 16 tests pass
- [ ] Manual: add an `hs300` driver, verify 6 outlets are created automatically
- [ ] Manual: click "Provision" on an existing driver, verify missing connectors are added

Fixes #1899

### DIFF
--- a/controller/device_manager/device_manager.go
+++ b/controller/device_manager/device_manager.go
@@ -1,10 +1,12 @@
 package device_manager
 
 import (
+	"fmt"
 	"log"
 	"strings"
 
 	"github.com/gorilla/mux"
+	"github.com/reef-pi/hal"
 	"github.com/reef-pi/rpi/i2c"
 
 	"github.com/shirou/gopsutil/v4/host"
@@ -14,6 +16,8 @@ import (
 	"github.com/reef-pi/reef-pi/controller/settings"
 	"github.com/reef-pi/reef-pi/controller/storage"
 	"github.com/reef-pi/reef-pi/controller/telemetry"
+	"github.com/reef-pi/reef-pi/controller/utils"
+	"net/http"
 )
 
 type DeviceManager struct {
@@ -53,7 +57,7 @@ func New(s settings.Settings, store storage.Store, t telemetry.Telemetry) *Devic
 	}
 	log.Println("device-manager subsystem initialized with", drvrs.Size(), "drivers")
 
-	return &DeviceManager{
+	dm := &DeviceManager{
 		bus:       bus,
 		drivers:   drvrs,
 		jacks:     connectors.NewJacks(drvrs, store),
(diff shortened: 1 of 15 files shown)


ranjib:
@copilot fix build failures due to translation checks

PR #2542, opened 2026-04-11
fix: check in capabilities snapshot and unblock gitignore

### LINKED ISSUE #1849: Capabilities test uses a snapshot that's not checked into source control
[front-end/src/configuration/capabilities.test.js](https://github.com/reef-pi/reef-pi/blob/main/front-end/src/configuration/capabilities.test.js) has:

```
describe('render capabilities component', () => {
  it('<Capabilities />', () => {
    const component = renderer.create(
      <Capabilities capabilities={[]} update={() => {}} />
    )

    const tree = component.toJSON()
    expect(tree).toMatchSnapshot()
  })
})
```

That tests that the capabilities view doesn't change. However, the actual snapshot isn't checked in anywhere, so it's just checking if it's changed since the last time you locally ran the test. It also would be a no-op in source control, because the snapshot doesn't exist so jest will on-the-fly generate it using the results of the test eval, effectively meaning it's a tautology.

Just noting this as a FYI. It's not blocking me on anything, since I just worked around the local failure by `rm -rf ./front-end/src/configuration/__snapshots__` to cause my local to regen it.

If you'd like to keep this test, I am happy to regen it and commit the snapshot.

ranjib:
## Summary

- Removes `__snapshots__` from `.gitignore` so Jest snapshot files can be committed
- Commits the existing `capabilities.test.js.snap` baseline

Without this fix, `capabilities.test.js` is a tautology: Jest generates the snapshot on first run and always passes because there is no checked-in baseline to compare against. Any developer deleting their local `__snapshots__/` directory (or CI) would silently skip the comparison.

## Test plan

- [ ] Run `cd front-end && npm test -- --testPathPattern=capabilities` — test should pass using the committed snapshot
- [ ] Modify `capabilities.jsx` (add/remove a capability) — test should now **fail**, confirming the snapshot is acting as a real guard

Fixes #1849


ranjib:
@copilot can you fix the build

PR #2551, opened 2026-04-12
feat: add cyclic PWM profile for wave-maker control

### LINKED ISSUE #1240: PWM Cyclic profile
I would like to be able to use Reef Pi to directly control a DC powerhead in a 'wave maker' pattern

To be able to do this it would need to be able to ramp up / down between two set values for a set duration.

I propose an additional profile within the PWM (Lighting) control, that provides this.

so essentially:

- Min value (percent)
- Max value (percent)
- Cycle Duration (time in seconds)

Assuming that people want to run multiple power heads it might also be good to include a phase shift value. This would offset the cycle start by the prescribed amount. A simple percentage (of duration) value would work

- Phase shift (percent)  

There are also various ramp shapes used in wave makers to modify the rate of change (acceleration rate / acceleration profile) this too could optionally be implemented, although I have found that regular sinusoidal waveforms are generally good enough and more in tune with a natural pattern.

ranjib:
## Summary

- Adds a new `cyclic` PWM profile to `controller/pwm_profile/` that continuously oscillates between `min` and `max` using a full sine wave based on elapsed Unix time
- Unlike the existing `sine` profile (bounded by wall-clock start/end times), the cyclic profile repeats indefinitely — no dead zone
- Parameters: **period** (seconds) and **phase_shift** (percent of period, for running multiple powerheads out of phase)
- Wires the profile into the factory, lighting UI profile selector, profile form, channel defaults, and Yup validation schema

## Formula

```
value = min + (max − min) × (sin(2π × elapsed / period) + 1) / 2
```

where `elapsed = (unix_time − phase_offset_secs) mod period`.

## Test plan

- [ ] `go test ./controller/pwm_profile/...` — all tests pass
- [ ] In the lighting UI, select a channel → choose **Cyclic** profile → enter period (e.g. 60) and phase shift (e.g. 0) → save
- [ ] Observe the channel value cycling between min and max over the configured period
- [ ] Create two lights/jacks with the same period but phase_shift 50 — confirm they are 180° out of phase

Closes #1240


ranjib:
@copilot fix the jest failure and translation check issues

PR #2565, opened 2026-04-12
feat: add mL/dose configuration for DC dosing pumps

### LINKED ISSUE #2008: Dosers Improvements
Hi @ranjib,
I'm trying the setup a Dosing pump in Reef-Pi and I think it'll be really nice to be able to setup the mL/dose instead of the duration to run the pump.
In this way, during the calibration, Reef-Pi can store how many seconds needs to dose X mL of liquid, then in the doser tab we can enter the mL that we want for each dose. (maybe with a small box that tell us how many mL we are dosing each day based on our schedule)
In this way on the dashboard we can track how many mL we have dose and (this will be very cool in my opinion) set an alarm based on the reservoir capacity left.
What do you think?
Best regards

Simone

ranjib:
Closes #2008

## Summary

- Adds a `VolumePerSecond` calibration constant to `DosingRegiment` (mL/s)
- Extends the calibration modal with a second step: after running the pump, the user enters the measured output volume in mL and clicks **Save calibration** — this calls the new `POST /api/doser/pumps/{id}/calibrate/save` endpoint which stores `volume / duration` as the mL/s rate
- The dosing runner automatically computes `duration = Volume / VolumePerSecond` when a pump is calibrated, so users can specify mL/dose instead of seconds
- The edit form switches the Duration field to a Volume (mL) field once a pump is calibrated, and shows the current rate below it

## Test plan

- [ ] Create a DC pump doser
- [ ] Open calibrate → set speed + duration → Run (pump runs)
- [ ] Enter measured mL in the new field → Save calibration
- [ ] Verify the edit form now shows Volume (mL) input instead of Duration
- [ ] Set a mL target → save → verify the pump runs for the correct computed duration
- [ ] Pump with `volume_per_second = 0` still uses plain Duration (no regression)


ranjib:
@copilot resolve merge conflicts

PR #2734, opened 2026-05-06
Avoid mutating AutoProfile state during render

ranjib:
## Summary
- remove duplicate AutoProfile constructor state initialization
- normalize slider values from a copied array during render instead of writing into state
- add regression coverage that sliderList leaves state.values unchanged

## Tests
- rtk yarn jest front-end/src/lighting/auto_profile.test.js
- rtk yarn standard front-end/src/lighting/auto_profile.jsx front-end/src/lighting/auto_profile.test.js
- rtk git diff --check


ranjib:
@copilot rebase this with latest main

PR #2738, opened 2026-05-06
Clone temperature schema fixtures before overrides

ranjib:
## Summary
- clone temperature schema test fixtures before per-case heater/cooler overrides
- preserve the base fixture between validations in the disabled heater/cooler test

## Validation
- yarn jest front-end/src/temperature/temperature_schema.test.js
- yarn standard
- git diff --check


ranjib:
@copilot rebase against main

PR #2740, opened 2026-05-06
Replace confirm Object.assign with spread

ranjib:
## Summary
- replace the remaining Object.assign in confirm utils with object spread
- add focused coverage for confirm option/message prop merging

## Validation
- rtk yarn jest front-end/src/utils/confirm.test.js
- rtk yarn standard
- rtk git diff --check


ranjib:
@copilot rebase against main

